PR #7, opened 2017-08-10
Add public_ipv4_subnet_size option to packet_device, plus acceptance test

t0mk:
This PR adds the options `public_ipv4_subnet_size` to the packet_device resource. The make it possible to allocate IPv4 subnet on device creation. More on this at
https://help.packet.net/technical/networking/custom-subnet-size

We had a user request for this.  

I had to update the packngo version.

I also wrote an acceptance test for the packet_device module. It takes several minutes, as the device provisioning takes a while.

I am open for comments and suggestions to improve this.

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -112,6 +112,12 @@ func resourcePacketDevice() *schema.Resource {
 				Optional: true,
 			},
 
+			"public_ipv4_subnet_size": &schema.Schema{
+				Type:     schema.TypeInt,
+				Optional: true,
+				ForceNew: true,
+			},
+
 			"tags": &schema.Schema{
 				Type:     schema.TypeList,
 				Optional: true,
@@ -125,12 +131,13 @@ func resourcePacketDeviceCreate(d *schema.ResourceData, meta interface{}) error 
 	client := meta.(*packngo.Client)
 
 	createRequest := &packngo.DeviceCreateRequest{
-		HostName:     d.Get("hostname").(string),
-		Plan:         d.Get("plan").(string),
-		Facility:     d.Get("facility").(string),
-		OS:           d.Get("operating_system").(string),
-		BillingCycle: d.Get("billing_cycle").(string),
-		ProjectID:    d.Get("project_id").(string),
+		HostName:             d.Get("hostname").(string),
+		Plan:                 d.Get("plan").(string),
+		Facility:             d.Get("facility").(string),
+		OS:                   d.Get("operating_system").(string),
+		BillingCycle:         d.Get("billing_cycle").(string),
+		ProjectID:            d.Get("project_id").(string),
+		PublicIPv4SubnetSize: d.Get("public_ipv4_subnet_size").(int),
 	}
 
 	if attr, ok := d.GetOk("user_data"); ok {
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -114,6 +114,7 @@ func resourcePacketDevice() *schema.Resource {
 
 			"public_ipv4_subnet_size": &schema.Schema{
 				Type:     schema.TypeInt,
+				Computed: true,
 				Optional: true,
 				ForceNew: true,
 			},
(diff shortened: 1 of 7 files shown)


packet/resource_packet_device.go
```diff
@@ -125,12 +130,13 @@ func resourcePacketDeviceCreate(d *schema.ResourceData, meta interface{}) error
...
-		OS:           d.Get("operating_system").(string),
-		BillingCycle: d.Get("billing_cycle").(string),
-		ProjectID:    d.Get("project_id").(string),
+		HostName:             d.Get("hostname").(string),
+		Plan:                 d.Get("plan").(string),
+		Facility:             d.Get("facility").(string),
+		OS:                   d.Get("operating_system").(string),
+		BillingCycle:         d.Get("billing_cycle").(string),
+		ProjectID:            d.Get("project_id").(string),
+		PublicIPv4SubnetSize: d.Get("public_ipv4_subnet_size").(int),
```
radeksimko: This field is optional, so if it's not defined it will come back as `0` - can you confirm this is in line with what the API expects? i.e. zero == undefined? If not, then we should probably use something like this instead:

```go
if v, ok := d.GetOk("public_ipv4_subnet_size"); ok {
	createRequest.PublicIPv4SubnetSize = v.(int)
}
```

t0mk:
@radeksimko Thanks for the review, it was a good point with the `public_ipv4_subnet_size`. It made me realize I mishandled it in the golang lib (my fault). I fixed it now (https://github.com/packethost/packngo/pull/25), so I will wait for the merge.

After the merge, the default zero will be handled alright in the packet golang lib. I will then update the govendor again, and write tests for the default case, and for subnet request.

t0mk:
@radeksimko, I addded another test for omitting the `public_ipv4_subnet_size` to packet/resource_packet_device.go. Can you please review the latest version of the device_test file?

This is not ready to merge yet, because I am still waiting for the merge of https://github.com/packethost/packngo/pull/25, which fixes handling of default value for `public_ipv4_subnet_size`.

Once the packngo PR will be merged, I will update vendor, squash these commits, rebase and ping you again.

Thanks!

packet/resource_packet_device_test.go
```diff
@@ -0,0 +1,152 @@
...
+
+		*device = *foundDevice
+
+		return nil
+	}
+}
+
+var testAccCheckPacketDeviceConfig_basic = fmt.Sprintf(`
+resource "packet_project" "terraform_test_project" {
+    name = "TerraformTestProject"
```
radeksimko: As we run most acceptance tests in parallel, static names like this one may cause issues - do you mind randomizing it? See how we do it in https://github.com/terraform-providers/terraform-provider-packet/blob/master/packet/resource_packet_volume_test.go#L82

packet/resource_packet_device_test.go
```diff
@@ -0,0 +1,152 @@
...
+  plan             = "baremetal_0"
+  facility         = "sjc1"
+  operating_system = "ubuntu_16_04"
+  billing_cycle    = "hourly"
+  project_id       = "${packet_project.terraform_test_project.id}"
+}`)
+
+var testAccCheckPacketDeviceConfig_request_subnet = fmt.Sprintf(`
+resource "packet_project" "terraform_test_project" {
+    name = "TerraformTestProject"
```
radeksimko: As above - ideally anyone should be able to run these two tests both in parallel without any issues.

t0mk:
Hi @radeksimko, the `public_ipv4_subnet_size` is more of a parameter of the device create request, than an attribute of the resource. From API point of view, It's actually not present in the device resource JSON, it's only in the creating POST request, that's why I didn't put it in the Terraform resource attribute. Of course I can dig it from the list of IP addresses (it's the CIDR prefix of the public IPv4), but I don't see the purpose of it.

It's not possible to update it for existing device, so forcenew: true applies, I will fix it. I will also put random suffix to the resource names in the test.

packet/resource_packet_device_test.go
t0mk: In Packet it's possible to have resources with same name, as they will have different UUID. Terraform uses ID to refer to the resoruces (just like it's in the test configs) so adding random suffices to the names doesn't really make sense, but I'll add it anyway.

t0mk:
I re-organized the commits and rebased on latest master.

t0mk:
Hi @radeksimko, I made the requested changes: The `public_ipv4_subnet_size` field is not updatable, so I added ForceNew:true, and I also read it in resourcePacketDeviceRead. 

If you think it's OK, please merge. I just run the tests with Packet and they run OK (with 01113de in), so it's a shipit from me!

radeksimko:
Thank you for making those changes, this now LGTM.

PR #10, opened 2017-08-23
packet_device: Add PXE arguments

### LINKED ISSUE #3: always_ipxe not supported
Packet [API](https://www.packet.net/developers/api/devices/) has a nifty ```"always_pxe": "boolean"```

```json
{
	"facility": "string",
	"plan": "string",
	"hostname": "string",
	"description": "string",
	"billing_cycle": "string",
	"operating_system": "string",
	"always_pxe": "boolean",
	"ipxe_script_url": "string",
	"userdata": "string",
	"locked": "boolean",
	"spot_instance": "string",
	"tags": [
		"string"
	],
	"project_ssh_keys": [
		"string"
	],
	"user_ssh_keys": [
		"string"
	],
	"features": [
		"string"
	]
}
```
Any chance this will be implemented in terraform? Right now i'm forced to make an additional local-exec provisioner, that updates the API to enable the always_ipxe, it seems rather convoluted for something so simple...

bzub:
Fixes #3

### DIFF
--- a/packet/resource_packet_device_test.go
+++ b/packet/resource_packet_device_test.go
@@ -27,6 +27,10 @@ func TestAccPacketDevice_Basic(t *testing.T) {
 					testAccCheckPacketDeviceAttributes(&device),
 					resource.TestCheckResourceAttr(
 						r, "public_ipv4_subnet_size", "31"),
+					resource.TestCheckResourceAttr(
+						r, "ipxe_script_url", ""),
+					resource.TestCheckResourceAttr(
+						r, "always_pxe", "false"),
 				),
 			},
 		},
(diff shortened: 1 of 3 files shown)


t0mk:
Hey @bzub, thanks to the PR! There is indeed demand for iPXE boot of packet devices.

In the tests, the operating system is set to "ubuntu_16_04", shouldn't it be "custom_ipxe"?

Can you clarify what's exactly going to happen when I set OS to "ubuntu_16_04" and will set "ipxe_script_url"? I mean will it actually boot ubuntu or the pxe?

I understand that testing what actually runs on that device is beyond the scope of this resources, but it would be good to be clear on that.

t0mk:
Maybe throw an error when 
- (ipxe_script_url is set && os != "custom_ipxe")
- (ipxe_script_url is not set && os == "custom_ipxe")
Those could be typos, and would be nice to catch by the provider.

bzub:
I'll fix the test OS, good catch.

Putting a condition on the OS is something I thought about, and I wasn't sure if that validation should go here or left to the Packet API. I agree it would be good to help a user that accidently put the wrong OS, so I'll add that condition. Thanks!

bzub:
I haven't tested setting non-custom_ipxe + ipxe_script_url. The API does disallow device creation with a useful error when you set an invalid/unresolvable ipxe_script_url, so in this case it probably errors or ignores the script url.

packet/resource_packet_device_test.go
```diff
@@ -136,3 +184,33 @@ resource "packet_device" "test_subnet_29" {
...
+}
+
+resource "packet_device" "test_always_pxe" {
+  hostname         = "test_always_pxe"
+  plan             = "baremetal_0"
+  facility         = "sjc1"
+  operating_system = "custom_ipxe"
+  billing_cycle    = "hourly"
+  project_id       = "${packet_project.test.id}"
+  always_pxe       = true
```
t0mk: If always_pxe is set to true, you have to supply "ipxe_script_url" params, like in the previous test. Otherwise, this acceptance test fails.

packet/resource_packet_device_test.go
```diff
@@ -136,3 +184,33 @@ resource "packet_device" "test_subnet_29" {
...
   public_ipv4_subnet_size = 29
 }`
+
+var testAccCheckPacketDeviceConfig_ipxe_script_url = `
+resource "packet_project" "test" {
+    name = "TerraformTestProject-%s"
+}
+
+resource "packet_device" "test_ipxe_script_url" {
+  hostname         = "test_ipxe_script_url"
```
t0mk: Also, please don't put underscores to hostnames, it's not allowed in packet.

packet/resource_packet_device_test.go
```diff
@@ -136,3 +184,33 @@ resource "packet_device" "test_subnet_29" {
...
+  ipxe_script_url  = "https://boot.netboot.xyz"
+}`
+
+var testAccCheckPacketDeviceConfig_always_pxe = `
+resource "packet_project" "test" {
+    name = "TerraformTestProject-%s"
+}
+
+resource "packet_device" "test_always_pxe" {
+  hostname         = "test_always_pxe"
```
t0mk: Dtto, please don't put underscores to hostnames, it's not allowed in packet.

t0mk:
Running acceptance tests now. If they pass, I will merge.

t0mk:
Acceptance tests pass, merging. Thanks for the PR!

PR #13, opened 2017-09-01
Set device user_data field to Sensitive

### LINKED ISSUE #12: Set packet_device's user_data field to Sensitive
### Terraform Version
Terraform v0.10.2

### Affected Resource(s)
- packet_device.example.user_data

### Output
Example: https://gist.github.com/bzub/c8c0c9587bc6e06a05129bceb947ce4b

I think the packet provider should set the user_data field to be Sensitive because it is often a very long string and contain secrets like private keys. Also, in the example I linked at least, it's not really human readable nor machine readable.

https://godoc.org/github.com/hashicorp/terraform/helper/schema#Schema

bzub:
This causes user_data to be omitted from log/console output.

Fixes #12

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -108,8 +108,9 @@ func resourcePacketDevice() *schema.Resource {
 			},
 
 			"user_data": &schema.Schema{
-				Type:     schema.TypeString,
-				Optional: true,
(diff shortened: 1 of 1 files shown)


t0mk:
Acceptance tests pass, I am merging

PR #15, opened 2017-09-01
Add field for root password

t0mk:
This PR make it possible to read the temporary root password on new device.

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -60,6 +60,12 @@ func resourcePacketDevice() *schema.Resource {
 				Computed: true,
 			},
 
+			"root_password": &schema.Schema{
+				Type:      schema.TypeString,
+				Computed:  true,
+				Sensitive: true,
+			},
+
 			"locked": &schema.Schema{
 				Type:     schema.TypeBool,
 				Computed: true,
@@ -244,6 +250,7 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("updated", device.Updated)
 	d.Set("ipxe_script_url", device.IPXEScriptURL)
 	d.Set("always_pxe", device.AlwaysPXE)
+	d.Set("root_password", device.RootPassword)
 
 	if len(device.HardwareReservation.Href) > 0 {
 		d.Set("hardware_reservation_id", path.Base(device.HardwareReservation.Href))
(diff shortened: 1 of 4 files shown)


packet/resource_packet_device.go
```diff
@@ -233,6 +239,7 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("updated", device.Updated)
 	d.Set("ipxe_script_url", device.IPXEScriptURL)
 	d.Set("always_pxe", device.AlwaysPXE)
+	d.Set("root_password", device.RootPassword)
```
radeksimko: Do I understand correctly that the API will _not_ return the password after 24 hours, so any `plan` or `apply` will wipe it from the state - or does it just become disabled on Packet's side?

If it's the former then I'm worried that this may be breaking for some users which reference the field elsewhere as it would "suddenly" become empty and affect other resources. Assuming the password cannot be changed maybe we could do something like this?

```
if d.IsNewResource() {
	d.Set("root_password", device.RootPassword)
}
```
so that it's set only during the creation and never wiped.

t0mk: The Packet Host infra logic will actively remove the root pass from /etc/shadow from the device (or do something similar), so that it's not possible to login there as root with this password after the 24 hours.

As for the API, after the 24 hours, the reply to the query for the device, will not contain the root_password field - due to the Device struct in packngo, the "root_password" in the TF resource will be set to an empty string.

I a way, it's a computed user-read-only attribute that changes with time, a bit like "uptime" attribute (if there would be any in the API). I don't really understand what it causes in eventual references.

Considering this, do you want me to set the root_password only on `d.IsNewResource()`?

radeksimko: Thanks for the explanation, that's helpful.

> a bit like "uptime" attribute (if there would be any in the API)

that's also why we usually avoid exposing such attributes like uptime, or dynamic timestamps generally.

> I don't really understand what it causes in eventual references.

Theoretically the user may reference it elsewhere, e.g. in https://www.terraform.io/docs/providers/vault/r/generic_secret.html and the current behaviour would cause the secret to be wiped.

It's questionable whether that's expected or unexpected behaviour 🤔 You could argue that the password should not be present when it's not valid anymore. 🤷‍♂️ 

I don't have too strong opinion about this, so I'll leave it up to you. Merge at will.

t0mk: Thanks for reflection, I'm a TF noob and don't see all the aspects yet, so your comments are really appreciated.

I do argue that password should be unset when it's not valid anymore, and this is what the current code does IMO.

Anyway, I merge, this is a small issue and can be later resolved easily.

PR #17, opened 2017-09-01
Fix iPXE Related Validations

### LINKED ISSUE #16: Should allow custom_ipxe OS with user_data and no ipxe_script_url
I noticed https://github.com/packethost/packngo/pull/21 and verified packet.net allows you to omit `ipxe_script_url` when `user_data` is specified along with `custom_ipxe` OS.

Current code does not allow `ipxe_script_url` to be empty when `custom_ipxe` is the OS. So the validation needs to be updated for the case above.

bzub:
Fixes #16

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -128,8 +128,9 @@ func resourcePacketDevice() *schema.Resource {
 			},
 
 			"ipxe_script_url": &schema.Schema{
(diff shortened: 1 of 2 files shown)


bzub:
Ready for review and acceptance tests!

t0mk:
I run acceptance tests, the new ones, and the Basic one. All good :+1:

PR #18, opened 2017-09-02
Add first-level IP address variables

### LINKED ISSUE #11: Document obtaining network information
### Affected Resource(s)
- packet_device

### References
https://www.terraform.io/docs/configuration/interpolation.html#user-map-variables
https://www.terraform.io/docs/configuration/interpolation.html#lookup-map-key-default-

Getting network information from packet_device.*.network is awkward due to limitations in Terraform around nested lists. Also, to get public/private/ipv4/ipv6 network elements you either have to assume the list is always ordered the same way (seems to work for me), or do complex lookups inside the nested list to verify values.

Maybe worth making the information available in a flat list or other, but for now I think obtaining for example the public ipv4 of a device should be documented.

Here's how I do it for public ipv4:
```hcl
lookup(packet_device.test_device.network[0], "address")
```

bzub:
- access_public_ipv6
- access_public_ipv4
- access_private_ipv4

These are copied from the "network" list of maps for convenience.

Fixes #11

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -71,6 +71,21 @@ func resourcePacketDevice() *schema.Resource {
 				Computed: true,
 			},
 
+			"access_public_ipv6": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"access_public_ipv4": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
+			"access_private_ipv4": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+
 			"network": &schema.Schema{
 				Type:     schema.TypeList,
 				Computed: true,
@@ -280,9 +295,19 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 		}
 		networks = append(networks, network)
 
-		if ip.AddressFamily == 4 && ip.Public == true {
-			host = ip.Address
-			ipv4SubnetSize = ip.CIDR
+		// Initial device IPs are fixed and marked as "Management"
+		if ip.Management {
+			if ip.AddressFamily == 4 {
+				if ip.Public {
+					host = ip.Address
+					ipv4SubnetSize = ip.CIDR
+					d.Set("access_public_ipv4", ip.Address)
+				} else {
+					d.Set("access_private_ipv4", ip.Address)
+				}
+			} else {
+				d.Set("access_public_ipv6", ip.Address)
+			}
 		}
 	}
 	d.Set("network", networks)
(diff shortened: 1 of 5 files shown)


bzub:
Ready for review and acceptance tests!

packet/resource_packet_device.go
```diff
@@ -64,6 +64,24 @@ func resourcePacketDevice() *schema.Resource {
 				Computed: true,
 			},
 
+			"access_public_ipv6": &schema.Schema{
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: true,
```
t0mk: Is the ForceNew necessary in computed fields? Maybe @radeksimko could clarify.

packet/resource_packet_device.go
```diff
@@ -255,9 +273,16 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 		}
 		networks = append(networks, network)
 
-		if ip.AddressFamily == 4 && ip.Public == true {
-			host = ip.Address
-			ipv4SubnetSize = ip.CIDR
+		if ip.AddressFamily == 4 {
```
t0mk: I noticed that the list of the addresses in the ip_addresses field of new device resource in the API is [public_IPv4 , IPv6, private_IPv4]. And when you assign flaoting IP, the order is [public_IPv4, flaoting_public_IPv4 ,IPv6, private_IPv4]. Based on my little tryout, the floating IP goes after the default public IPv4.

If floatign IPv4 is assigned to device, and then access_public_ipv4 is read this way (as in the current code), the floating IP will rewrite the defualt public IPv4, because it comes later in the list of the ip_addresses. I think this is not desired behavior.

Could you change it so that the "access_public_ipv4" is read only once (i.e. is not rewritten if there is a floating IPv4 on this device)?

t0mk:
I run the basic acceptance test, and beside my comments in the review it it's alright.

t0mk:
Blocked by https://github.com/packethost/packngo/pull/40

@bzub please review the linekd packngo PR if you have time.

packet/resource_packet_device.go
radeksimko: Fields can be both `Computed` and `Optional` in which case they may _also_ be `ForceNew`, but for Computed-only field can't really be `ForceNew`, because such field never occurs in the config, so there's nothing to trigger the recreation process (API changing computed values doesn't isn't interpreted as change, it's just simply written to state and updated wherever it's referenced).

Thanks for pointing this out - I'll see if we can add this to our schema validation logic
https://github.com/hashicorp/terraform/blob/cbb512d374a80d9d2c7cd49ca0644c576ab2335f/helper/schema/schema.go#L548

bzub:
Please take another look, thanks!

PR #19, opened 2017-09-04
vendor: github.com/hashicorp/terraform/...@v0.10.3

radeksimko:
Here I'm bumping vendored terraform to get test for `id` fields which are unnecessary and became reserved in https://github.com/hashicorp/terraform/pull/15695

I had to bump `go-plugin` because govendor doesn't deal with transitive dependencies.


radeksimko:
Blocked by #20

t0mk:
I installed go 1.9, and then run the basic acceptance test and it seems to work alright. :+1:

PR #20, opened 2017-09-04
travis: Bump Go to 1.9

radeksimko:
This is necessary before bumping vendored `hashicorp/terraform` to `0.10.3` as it uses some `1.9` features.


t0mk:
To review this, I would only install Go 1.9 and try if it compiles. As the travis run de facto did that, I :+1:. this.

PR #26, opened 2017-09-14
Enable next-available hw resevation in device resource

t0mk:
This makes it possible to provision device on reserved hardware without explicitly specifying the UUID of the hw reservation. Packet has customer demand for this.

In Packet this is done by passing `next-available` instead of the hw reservation ID when creating a device. The created device then has the UUID of the used hw reservation in the resource attributes.  

I realize it's a bit terraform-unfriendly, and that's why I had to specify the DiffSuppressFunc. The device creation did not change, but the behavior on update is:
If new hwr\_id (value from tf file) is "next-available" and old hwr\_id (value from API read) is non-empty; then suppress diff (i.e. do skip, even though this "ForceNew" field would normally cause resource update).

It's alright this way I think - if "next-available" is in the hwr\_id field, and computed hwr\_id from the API is non-empty, we don't want to do anyhting.

This would work for import too I think.

@radeksimko, please check this out, even though I think this PR will not make you very happy.

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -157,7 +157,14 @@ func resourcePacketDevice() *schema.Resource {
 			"hardware_reservation_id": &schema.Schema{
 				Type:     schema.TypeString,
 				Optional: true,
+				Computed: true,
 				ForceNew: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					if new == "next-available" && len(old) > 0 {
+						return true
+					}
+					return false
+				},
 			},
 
 			"tags": &schema.Schema{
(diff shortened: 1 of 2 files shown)


t0mk:
Thanks for the comment. It's not easy to manage this kind of pool of limited resources with Terraform.

I don't know exactly how data sources behave, but, in your example (considering there are at least 2 free hw reservations), on terraform init, the device will be created, and on subsequent refresh, there will be a diff for the `hardware_reservation_id` param, because the next-available will be the next free reservation. So I think it doesn't solve the break of the idempotency, which is quite crucial for TF.

It's possible to implement next-available in the client logic as you suggested (based on the `provisionable` param of the reservations in the API), but I'm afraid that would misbehave when creating more than one device at a time, or with count parameter. This way, the assignment of hw reservation is left to the Packet API, leaving TF provider thinner.

t0mk:
Thanks. If this causes any trouble, I'll revert. Also if there's some more reasonable workaround, I'm happy to review.

PR #27, opened 2017-09-24
Device spot market support

bzub:
New acceptance tests pass. I'm going to take it for a spin with my TF modules but should be ready for review.

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -561,7 +561,7 @@ func floatAtLeast(min float64) schema.SchemaValidateFunc {
 		}
 
 		if v < min {
-			errs = append(errs, fmt.Errorf("expected %s to be at least (%d), got %d", k, min, v))
+			errs = append(errs, fmt.Errorf("expected %s to be at least (%f), got %f", k, min, v))
 			return
 		}
 
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -11,6 +11,17 @@ import (
 	"github.com/packethost/packngo"
 )
 
+const (
+	terminationTimeRoundVal = time.Second * 10
+)
+
+type timeParserFunc func(s string) (time.Time, error)
+
+var terminationTimeParsers = []timeParserFunc{
+	timeFromRFC3339,
+	timeAfterDuration,
+}
+
 func resourcePacketDevice() *schema.Resource {
 	return &schema.Resource{
 		Create: resourcePacketDeviceCreate,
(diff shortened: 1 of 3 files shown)


bzub:
Works well in my tests.

bzub:
One thing, there seems to be an issue when a device is deleted at `termination_time`. Any terraform command that refreshes state errors out due to 403 forbidden, and requires manual intervention to fix (delete or taint the resource). I'll see if I can fix that quick.

t0mk:
I guess the same error will happen when devices is shut down because market price had risen over the spot_price_max?

Maybe look at how it's handled elsewhere? In AWS, they have a different resource for spot instances:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_instance_request.go

I'm not sure if there any other spot computes in TF.

bzub:
The weird thing is that one device seems to be marked as deleted as expected, while the other gets 403 forbidden from packet API. Same termination time for both. Might be a packet api fluke.

bzub:
Or... we're querying the API wrong and need to list devices under the packet project ID first, and only look up devices from that list... I could see this happening if a device is reprovisioned by another packet user and the state hasn't been refreshed since the device was deleted.

For now, I think I'll check for 403 forbidden, then check if we're after `termination_time`, and if both are true assume the device is deleted and delete from state. This is conservative to avoid an issue where permission to a device is temporarily not allowed, but then restored, in which case the device shouldn't be removed from TF state but the error should prevent TF operations.

bzub:
Arg, my thoughts were not based on reality somewhat. I don't think there's device level permissions. We might have to just make sure the device isn't in the project list. Actually simpler that way.

bzub:
Should be good to go now pending review :)

t0mk:
This is quite a big PR, and the persistence of the spot device is problematic from the Terraform state point of view I think.

Maybe @radeksimko could take a look and give some directions on how to handle spot devices, which can be stopped when spot price goes over what the user wants to pay?

radeksimko:
Thank you for the contribution @bzub 

I agree with @t0mk that the persistence is problematic in the way Terraform manages resources and the way user expects Terraform to do it.

The main reason it's problematic is because of Packet's API behaviour (that's not to diss Packet, but to explain the situation). AWS is different and resource also works differently - it is called `_request`. The request is basically an abstraction layer above EC2 instances themselves and it exists in the API regardless of instance status (or market price), and it can be cancelled at any time.

> For now, I think I'll check for 403 forbidden, then check if we're after termination_time, and if both are true assume the device is deleted and delete from state.

That's not how most Terraform users would expect any resource to behave. With that logic we could also support short-lived resources, like database records, jobs in a queue etc. - which is IMO a slippery slope. Terraform was not designed to handle such dynamic & short-lived resources in this context.

If we can make Packet change the API to work in a similar way AWS EC2 works, I'd be happy to accept a contribution, but I'm instead inclined to close this one as I feel it would cause too much confusion for users.

I hope the explanation makes sense.

t0mk:
That's too bad. To be fair, TF supports "preemptible" flag in the google compute instance resource, and that does some very strange things too:
https://github.com/terraform-providers/terraform-provider-google/blob/master/google/resource_compute_instance.go#L1306

bzub:
> For now, I think I'll check for 403 forbidden, then check if we're after termination_time, and if both are true assume the device is deleted and delete from state.

FYI that is not the approach I ended up on. Instead of assuming anything, I check the project for the device being refreshed, and if it's not there then it's marked successfully gone. If the config is still there for the device, naturally TF will want to recreate it. Seems like expected behavior if anyone chooses to use `termination_time`.

nathangoulding:
I'm late to the party here, but AFAIK EC2 has a spot market "request" that lives separately from individual spot instances, and this entity influences whether more instances are created automatically when certain metrics move into or out of predefined thresholds. We don't have support for this construct in our API, although something similar is planned as part of broader enhancements to our spot market and related tooling.

To keep things simple, a spot device in Packet is just like any another device. These devices may get deleted externally via the portal/API by another user, or in this case due to the current market price in that type/facility.

radeksimko:
Hi @nathangoulding 
this is pretty much the same conclusion I came to.

> something similar is planned as part of broader enhancements to our spot market and related tooling.

That's great.

@bzub Are you ok with closing this PR and reimplementing it once Packet comes up with those mentioned enhancements?

bzub:
Yup we will revisit when Packet enhances the API. Thanks, sorry for the delay @radeksimko

nathangoulding:
This is [now supported](https://www.packet.net/resources/changelog/persistent-spot-market-bid/) in Packet's API.

PR #28, opened 2017-09-25
Packngo bump

t0mk:
Bumping packngo to newest version adding
- volume attachments
- floating ip subnet handling
- HTTP unauthorized error message
- spot instance test


bzub:
LGTM

PR #35, opened 2017-10-20
support for Custom Partitioning and RAID

### LINKED ISSUE #34: Custom partitioning and RAID support
In Packet it's possible to define custom partitioning and RAID config in a device creation parameter:
https://help.packet.net/technical/storage/custom-partitioning-raid

It would be nice to have this in TF provider.

The `storage` parameter format is (at first glance) the same as in CoreOS Ignition, supporting disks, filesystems and raid types:

```
{
 "disks":<array of subobjects>,
 "filesystems":<array of subobjects>,
 "raid":<array of subobjects>, <- optional
}
```

As this is nicely structured, it could be done in a more sophisticated way than just JSON string.

There's already a TF provider for Ignition, so I will look on how the schema are defined there:
https://github.com/terraform-providers/terraform-provider-ignition

Thanks to @bzub for pointing me to Ignition (IMO it wouldn't hurt to mention it in the Packet CPR doc).

t0mk:
This PR adds a "storage" field to the device resource, where uses can specify custom partitioning and RAID setup. Packet has customer demand for this. The format of the specification is documented at https://help.packet.net/technical/storage/custom-partitioning-raid.

The storage specification is basically a JSON document specifying partitions, mountpoints and RAID setup; not unlike CoreOS ignition (https://coreos.com/ignition/docs/latest/configuration-v2_1.html), but not completely same IIUC. In Packet, also non-CoreOS devices can be partitioned.

The challenge here (from TF state point of view) was to make sure that only changes in the map structure will generate diff in TF state. Simply changing the order of how key-values are declared in the JSON document should not be detected as a change.

I normalize the JSON (json.Marshal(json.Unmarshal(jsonDoc)), so that the same maps are hopefully represented by the same JSON document. It's an approach I noticed in the aws provider. I do it through the StateFunc of the storage field.

There are no acceptance tests for this, because the `storage` can only be specified for device deployed on reserved hardware.

Fixes #34


t0mk:
@radeksimko Please take a look when you have time.

packet/provider.go
```diff
@@ -31,6 +34,33 @@ func Provider() terraform.ResourceProvider {
 	}
 }
 
+func normalizeJSON(jsonValue interface{}) (string, error) {
```
catsby: Terraform offers a `structure` package with many methods that look like this, we encourage (but do not require) you use that

- https://github.com/hashicorp/terraform/tree/master/helper/structure

packet/resource_packet_device.go
```diff
@@ -172,6 +173,23 @@ func resourcePacketDevice() *schema.Resource {
...
 			},
+			"storage": &schema.Schema{
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				StateFunc: func(v interface{}) string {
+					s, _ := normalizeJSON(v)
+					return s
+				},
+				ValidateFunc: func(v interface{}, k string) (ws []string, es []error) {
```
catsby: Terraform includes a package `validation` with a `ValidateJsonString` method you could use here

- https://github.com/hashicorp/terraform/blob/a41d5d634d282654a1c859ad4507463c25e82a00/helper/validation/validation.go#L163-L168

packet/resource_packet_device.go
```diff
@@ -276,6 +302,7 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	d.Set("ipxe_script_url", device.IPXEScriptURL)
 	d.Set("always_pxe", device.AlwaysPXE)
 	d.Set("root_password", device.RootPassword)
+	d.Set("storage", device.Storage)
```
catsby: For any non-scalar value (e.g. `[]string` or `map[string]interface{}`), you should check the error returned from `d.Set`:

```
if err := d.Set("storage", device.Storage); err != nil {
  log.Printf("[ERR] Error setting storage for (%s): %s", d.Id(), err)
}
```

packet/provider.go
t0mk: Great! Of course prefer to use inbuilts rather than new code. This taught me not to look to the aws provider for inspiration :D.

packet/resource_packet_device.go
t0mk: dtto

packet/resource_packet_device.go
t0mk: OK, will do.

t0mk:
@catsby could you please ACK my fixes to your review?

my commit that addresses your review is:
https://github.com/terraform-providers/terraform-provider-packet/pull/35/commits/595449fc027255f81662318298e62731463d960a

PR #37, opened 2017-10-23
Add datasource for pre-created (elastic) IP blocks

t0mk:
This PR adds a datasource for reserved IP blocks so that users can take advantage of the pre-created IPv6 and IPv4 blocks.

### DIFF
--- a/packet/resource_packet_ip_attachment.go
+++ b/packet/resource_packet_ip_attachment.go
@@ -69,7 +69,7 @@ func resourcePacketIPAttachmentRead(d *schema.ResourceData, meta interface{}) er
 
 	d.Set("device_id", path.Base(assignment.AssignedTo.Href))
 	d.Set("cidr_notation",
-		fmt.Sprintf("%s/%d", assignment.Address, assignment.CIDR))
+		fmt.Sprintf("%s/%d", assignment.Network, assignment.CIDR))
 
 	return nil
 }
--- a/packet/resource_packet_reserved_ip_block.go
+++ b/packet/resource_packet_reserved_ip_block.go
@@ -101,7 +101,6 @@ func resourcePacketReservedIPBlockCreate(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("Error reserving IP address block: %s", err)
 	}
 
-	d.Set("cidr_notation", blockAddr.Address)
 	d.Set("facility", facility)
 	d.Set("quantity", quantity)
 	d.Set("project_id", projectID)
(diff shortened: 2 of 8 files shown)


epiloque:
this is awesome! thank you so much @t0mk. 👏☺️
I’ll give it a try a little later!

t0mk:
This is ready for review.

packet/resource_packet_ip_attachment.go
```diff
@@ -69,7 +69,7 @@ func resourcePacketIPAttachmentRead(d *schema.ResourceData, meta interface{}) er
 
 	d.Set("device_id", path.Base(assignment.AssignedTo.Href))
 	d.Set("cidr_notation",
-		fmt.Sprintf("%s/%d", assignment.Address, assignment.CIDR))
+		fmt.Sprintf("%s/%d", assignment.Network, assignment.CIDR))
```
catsby: Are these two values `Address` and `Network` always equal? Is there possibility one is different, such that this change causes a drift or diff in upgrading?

packet/resource_packet_reserved_ip_block.go
```diff
@@ -131,9 +123,37 @@ func resourcePacketReservedIPBlockRead(d *schema.ResourceData, meta interface{})
...
+		// The following logic will hold as long as /64 is the smallest assignable subnet size.
+		bits := 64 - reservedBlock.CIDR
+		if bits > 30 {
+			return fmt.Errorf("Strange (too small) CIDR prefix: %d", reservedBlock.CIDR)
+		}
+		d.Set("quantity", 1<<uint(bits))
+	}
 	d.Set("project_id", path.Base(reservedBlock.Project.Href))
-	d.Set("cidr_notation", fmt.Sprintf("%s/%d", reservedBlock.Address, reservedBlock.CIDR))
+	d.Set("cidr_notation", fmt.Sprintf("%s/%d", reservedBlock.Network, reservedBlock.CIDR))
```
catsby: same `Address` -> `Network` question as before

packet/resource_packet_ip_attachment.go
t0mk: This is a bug fix, there should have been the Network since the beginning, so yes, this might theoretically cause diff when upgrading. But I think that "terraform refresh"  How shall I resolve this?

packet/resource_packet_reserved_ip_block.go
t0mk: dtto

t0mk:
@catsby thanks for review, the Address -> Network is a bugfix. I will not bore you with Packet API tech reasons, it was my fault in the first place. How shall I resolve it?

catsby:
The problem arises when someone is depending on `cidr_notation` down the chain, e.g. 

```
something = "${packet_reserved_ip_block.thing.cidr_notation}"
```

Depending on what that _down the chain_ thing is, the consequences could be dire. In my testing though, I didn't see much change... 

At the very least this needs to be marked with special attention in the CHANGELOG once it's merged. I'll let you decide if this warrants a 2.0.0 release due to breaking change.

t0mk:
Hi @catsby, you are right in the sense that it's a change worth-noting. However, I think not many people use the reserved block resource yet - it's a relatively new feature. I read semver, and I don't think this small change justifies 2.0.0 - it's not breaking, just might be a bit confusing.

I mentioned it in changelog.

Issue #47, opened 2018-02-20
Provider crashes when trying to rename ssh key


BRMatt:
### Terraform Version

```console
$ terraform -v
Terraform v0.11.3
+ provider.null v1.0.0
+ provider.packet v1.2.0
```

### Affected Resource(s)
- `packet_ssh_key`

### Terraform Configuration Files

```hcl
provider "packet" {
  auth_token = ""
}

resource "packet_ssh_key" "foobar" {
  name = "packer-foobar-hello"
  public_key = "${file("./ssh_keys/id_rsa.pub")}"
}
```

You can generate the ssh key using

```console
$ ssh-keygen -t rsa -C "used for sshing into ephemeral packet.net instances" -P '' -f ssh_keys/id_rsa
```

### Debug Output

Panic output is at the end of the gist

https://gist.github.com/BRMatt/23051b8816a2f75175f1f86fb48bea95

### Expected Behavior

The Packet provider should've renamed the ssh key, and updated the state file.

### Actual Behavior

The provider renamed the ssh key, but crashed before the state file could be be update.

### Steps to Reproduce

1. `terraform apply`
1. change the `name` attribute of the `packet_ssh_key`
1. `terraform apply` crashes

t0mk:
@BRMatt Thanks for the report. I am looking at this and I see what the culprit is. I will try to fix it by today.

t0mk:
OK, this goes deeper - to https://github.com/packethost/packngo

jasmingacic:
This seems to be fixed. Can we close this issue?

edevenport:
@BRMatt Can you verify this issue has been corrected now and close the ticket if it has? Thank you.

Issue #44, opened 2018-01-25
Packet provider crash


pgporada:
### Provider Version
1.2

### Terraform Version
0.11.2

### Affected Resource(s)
Please list the resources as a list, for example:
- packet_project
- packet_ssh_key

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
I'm following along from https://www.terraform.io/docs/providers/packet/index.html
```
$ cat main.tf 
variable "packet_api_key" {}
variable "packet_project_name" {}

terraform {
  required_version = ">= 0.11.2"
}

provider "packet" {
  auth_token = "${var.packet_api_key}"
}

module "packet_server" {
  source       = "./modules/packet.net/"
  project_name = "${var.packet_project_name}"
}

output "project_name" {
  value = "${var.packet_project_name}"
}

output "project_id" {
  value = "${module.packet_server.project_id}"
}
```

```
$ cat modules/packet.net/packet.net.tf 
variable "project_name" {}

resource "packet_project" "phil" {
  name = "${var.project_name}"
  payment_method = "test"          # Only required for a non-default payment method
  #payment_method = "Phil Was Here"          # Only required for a non-default payment method
}

resource "packet_ssh_key" "key_phil" {
  name       = "pgporada"
  public_key = "${file("${path.module}/../../ssh_keys/packet_pgporada.pub")}"
}

output "project_id" {
  value = "${packet_project.phil.id}"
}

output "project_payment_method" {
  value = "${packet_project.phil.payment_method}"
}
```

### Panic Output
https://gist.github.com/pgporada/c888812d378f8b45553cada9660954aa for payment method of "Phil Was Here"
https://gist.github.com/pgporada/2038ed44fcd62b77a72924a7cbcebbbe for payment method of "test"

I am unsure what the payment method should be if it's not the default. Additionally I get an error, but no crash, if I try to output the payment method when using the default.
```
$ terraform apply -var-file environments/test/test.tfvars
packet_ssh_key.key_phil: Refreshing state... (ID: efc554a2-1b1a-4248-abb6-0c1f7af334f5)

Error: Error running plan: 1 error(s) occurred:

* module.packet_server.output.project_payment_method: Resource 'packet_project.phil' does not have attribute 'payment_method' for variable 'packet_project.phil.payment_method'
```

### Expected Behavior
Not to see a crash

### Actual Behavior
[...]

PR #48, opened 2018-02-27
packngo bump

t0mk:
Use newer packngo, in order to fix
* https://github.com/terraform-providers/terraform-provider-packet/issues/47
* https://github.com/terraform-providers/terraform-provider-packet/issues/44


t0mk:
I verified that the acceptance tests pass in HashiCorp TeamCity.

PR #55, opened 2018-03-23
Add support for spot instances

yoink00:
Added support for 'spot_instance' and 'spot_price_max' from the Packet API


yoink00:
OK. Cool.

PR #61, opened 2018-04-25
Fix updates of volume resource

### LINKED ISSUE #60: Provider Packet doesn't update volume sizes
When a Packet volume size attribute is modified, Terraform indicates it will update the volume (with a ~) but no change happens.

Terraform will perform the following actions:

  ~ packet_volume.ETN_2
      size:      "15" => "25"

It appears the resourcePacketVolumeUpdate doesn't actually update the size. The Packet API does support volumes size updates. This function needs to be updated to include code for the size attribute.

https://github.com/terraform-providers/terraform-provider-packet/blob/fc2b8ee6de6dfb40ea6da1147c8f6ced4ef34744/packet/resource_packet_volume.go#L247-L265

The correct action would be for the size to be updated.

Sample output when changing the size from 100 to 222. It says it will change the item but the Packet GUI and the Terraform output still shows the old size.

packet_volume.volume1: Modifying... (ID: 6dd21377-9c3a-4d65-822e-cb0d92e3a959)
  size: "100" => "222"
packet_volume.volume1: Modifications complete after 1s (ID: 6dd21377-9c3a-4d65-822e-cb0d92e3a959)

Apply complete! Resources: 0 added, 1 changed, 0 destroyed.

Outputs:

Size = 100

studarus@shell:~/Volume$ ./terraform version
Terraform v0.11.7
+ provider.packet v1.2.2

t0mk:
This PR bumps packngo and fixes update of the volume resource.

Fixes #60.

### DIFF
--- a/packet/resource_packet_volume.go
+++ b/packet/resource_packet_volume.go
@@ -122,6 +122,7 @@ func resourcePacketVolumeCreate(d *schema.ResourceData, meta interface{}) error 
 		PlanID:     d.Get("plan").(string),
 		FacilityID: d.Get("facility").(string),
 		Size:       d.Get("size").(int),
+		Locked:     d.Get("locked").(bool),
 	}
 
 	if attr, ok := d.GetOk("billing_cycle"); ok {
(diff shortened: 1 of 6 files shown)


t0mk:
Tests pass, so I merge.

PR #63, opened 2018-04-25
Added Organization resource.

miry:
Allow create projects with organization id.

```
resource "packet_organization" "test" {
  name = "foobar"
  description = "test"
  website = "http://google.com"
  twitter = "miry_sof"
}

resource "packet_project" "foobar" {
  name = "foobar"
  organization_id = "${packet_organization.test.id}"
}
```

It is example how it could be done.

Another solution to move `packet_organization` under `provider` resource similar to https://www.terraform.io/docs/providers/aws/, with out create and update resources.

### DIFF
--- a/packet/provider.go
+++ b/packet/provider.go
@@ -24,6 +24,7 @@ func Provider() terraform.ResourceProvider {
 			"packet_device":            resourcePacketDevice(),
 			"packet_ssh_key":           resourcePacketSSHKey(),
 			"packet_project":           resourcePacketProject(),
+			"packet_organization":      resourcePacketOrganization(),
 			"packet_volume":            resourcePacketVolume(),
 			"packet_volume_attachment": resourcePacketVolumeAttachment(),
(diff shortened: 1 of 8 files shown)


t0mk:
Hi, thanks for the PR @miry.

I have a few requests:
- Can you please rebase on current master?
- Can you please rewite the Update method to sth like
https://github.com/terraform-providers/terraform-provider-packet/blob/master/packet/resource_packet_ssh_key.go#L90

The attributes of the UpdateRequest structs are pointers intentionaly, and the resource attrs that are not changed should not be set at all in the UpdateRequest struct (leaving them to `nil`, which will omit them from the json dict in the HTTP PATCH request).

miry:
@t0mk 

>Can you please rewrite the Update method to use d.HasChange()?

I tried to update project if organization id changed. But I could not change organization for project via API.

miry:
@t0mk It looks great. +1

PR #67, opened 2018-04-30
make: Add website + website-test targets

radeksimko:


t0mk:
I run make for the two targets, I could see the docs for packet provider on localhost, that's nice for debugging style, cool.

The website-test ran through, it succeeded though I don't know what it exactly did.

PR #68, opened 2018-05-25
Fix. Handle the case when an IP attachment is missing or has been deleted.

kwilczynski:
This commit adds a single check for 404 Not Found status code to ensure that
refreshing state of and/or deleting an IP assignment would not result in an
error should the resource could not be found (e.g., it has been already
deleted by other means, etc.).

Signed-off-by: Krzysztof Wilczynski <kw@linux.com>

### DIFF
--- a/packet/resource_packet_ip_attachment.go
+++ b/packet/resource_packet_ip_attachment.go
@@ -2,6 +2,7 @@ package packet
 
 import (
 	"fmt"
+	"log"
 	"path"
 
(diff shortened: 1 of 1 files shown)


kwilczynski:
Hi @tombuildsstuff, thank you! Sadly, I cannot merge this, so we need to wait for the maintainers, or otherwise.

t0mk:
test passes, I merge

PR #70, opened 2018-09-05
[datasource] packet_operating_system -> data source

jasmingacic:
Here is the new data source for packet provider.
```
make testacc TESTARGS='-run=TestAccPacketOperatingSystem_Basic'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run=TestAccPacketOperatingSystem_Basic -timeout 120m
?       github.com/terraform-providers/terraform-provider-packet        [no test files]
=== RUN   TestAccPacketOperatingSystem_Basic
--- PASS: TestAccPacketOperatingSystem_Basic (4.76s)
PASS
ok      github.com/terraform-providers/terraform-provider-packet/packet 4.788s
```

### DIFF
--- a/website/docs/d/operating_system.html.markdown
+++ b/website/docs/d/operating_system.html.markdown
@@ -1,9 +1,9 @@
 ---
 layout: "packet"
-page_title: "Packet: opearating_system"
+page_title: "Packet: operating_system"
 sidebar_current: "docs-packet-datasource-operating-system"
 description: |-
-  Get an operationg on a Packet Operating System image
+  Get a Packet operating system image
 ---
 
 # packet\_operating\_system
@@ -20,6 +20,15 @@ data "packet_operating_system" "example" {
   provisionable_on = "baremetal_1"
 }
 
+resource "packet_device" "server" {
+  hostname         = "tf.coreos2"
+  plan             = "baremetal_1"
+  facility         = "ewr1"
+  operating_system = "${data.packet_operating_system.example.id}"
+  billing_cycle    = "hourly"
+  project_id       = "${packet_project.cool_project.id}"
+}
+
 ```
 
 ## Argument Reference
@@ -31,5 +40,5 @@ data "packet_operating_system" "example" {
 
 ## Attributes Reference
 
- * `operating_system` - Opearting system of a device.
+ * `operating_system` - Operating system of a device.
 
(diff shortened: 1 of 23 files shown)


t0mk:
hi, can you illustrate a practical use case of the datasource in the markdown doc? So that users understand what's it good for? like in

https://www.terraform.io/docs/providers/do/d/image.html

website/docs/d/operating_system.html.markdown
```diff
@@ -0,0 +1,35 @@
+---
+layout: "packet"
+page_title: "Packet: opearating_system"
```
t0mk: s/opearating/operating

website/docs/d/operating_system.html.markdown
```diff
@@ -0,0 +1,35 @@
...
+sidebar_current: "docs-packet-datasource-operating-system"
+description: |-
+  Get an operationg on a Packet Operating System image
```
t0mk: description doesn't make sense

website/docs/d/operating_system.html.markdown
```diff
@@ -0,0 +1,35 @@
...
+## Argument Reference
+
+ * `distro` - (Optional) Name of the OS distribution.
+ * `name` - (Optional) Name or part of the name of the distribution. Case insensitive.
+ * `provisionable_on` - (Optional) Plan name.
+ * `version` - (Optional) Version of the distribution
+
+## Attributes Reference
+
+ * `operating_system` - Opearting system of a device.
```
t0mk: s/Opearting/Operating/

jasmingacic:
@t0mk here you go.

t0mk:
@jasmingacic Thanks! I tested the example and tried a few corner cases, and I confirm it works.

Just one more thing I suggest:
If I would do it, I would add a computed param for slug to the schema
```go
"slug": {
	Type:     schema.TypeString,
	Computed: true,
},
```

I can see that the slug goes to the `Id` of the resource, and another param would be duplicating it, but it might feel more consistent to the Packet users to use the `slug` param in hcl templates (` operating_system = "${data.packet_operating_system.example.slug}"`), instead of the Id, which might wrongly appear like UUID (and there's no UUID in the OS resources in Packet).

It's your PR so I leave that ^ up to you to decide. Let me know if you want to do it, and I will merge or wait accordingly.

jasmingacic:
I've added `slug` as a computed field. So if you think it is ready go ahead and merge it.

PR #71, opened 2018-09-10
Data source spot market price

jasmingacic:
@t0mk can you please take a look

This is one bit required for spot market requests to be fully implemented.

### DIFF
--- a/packet/datasource_spot_market_prices.go
+++ b/packet/datasource_spot_market_prices.go
@@ -1,0 +1,54 @@
+package packet
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourceSpotMarketPrices() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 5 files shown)


t0mk:
@jasmingacic thanks for the PR, I am taking a look.

I would change the file name to singular `datasource_packet_spot_market_price.go` so that it's the same as the name of the datasource.

packet/datasource_spot_market_prices.go
```diff
@@ -0,0 +1,54 @@
+package packet
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourceSpotMarketPrices() *schema.Resource {
```
t0mk: change to singular `dataSourceSpotMarketPrice`, as it only gets one price

t0mk:
I tested the datasource and it works. I understand that the lack of test is because it's nontrivial to check for a float64 attribute with the terraform testing suite.

PR #72, opened 2018-09-11
[New feature] Spot Market Request

jasmingacic:


jasmingacic:
@t0mk @radeksimko can one of you take a look at this?

t0mk:
I will take a look tomorrow.

t0mk:
why no test? It's unlikely to be OK with HashiCorp engs wihout one, and it's hard to evaluate. I suggest one test where devices_max > 1 and devices_min != devices_max.

Example of similar resource test:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request_test.go

website/docs/r/spot_market_request.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+  "max_bid_price" = 100.00
+  "facilities"    = ["ewr1"]
+  "devices_min"   = 1
+  "devices_max"   = 1
+
+  "instance_parameters" {
+    "hostname"         = "testspot"
+    "billing_cycle"    = "hourly"
+    "operating_system" = "coreos_stable"
+    "plan"             = "baremetal_1"
```
t0mk: maybe put `baremetal_0` to make copy&paste test more lightweight

website/docs/r/spot_market_request.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+Provides a Packet Spot Market Request resource to allow you to
+manage spot market requests on your account.
+
+## Example Usage
+
+```hcl
+# Create a spot market request
+resource "packet_spot_market_request" "req" {
+  project_id      = "${packet_project.cool_project.id}"
+  "max_bid_price" = 100.00
```
t0mk: maybe something like `0.03` for `baremetal_0` pricing to make copy&pasted TF template more real/usable

packet/resource_packet_spot_market_request.go
```diff
@@ -0,0 +1,223 @@
...
+
+func resourcePacketSpotMarketRequestRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*packngo.Client)
+
+	smr, _, err := client.SpotMarketRequests.Get(d.Id(), &packngo.ListOptions{Includes: "devices,facilities"})
+	if err != nil {
+		return err
+	}
+
+	deviceIDs := []string{}
```
t0mk: You want to 
`deviceIDs := make([]string, len(smr.Devices))`

otherwise you assign to empty slice

t0mk:
Add parameter `wait_for_fulfillment`, it might be useful for users who want to be sure that all the devices are created.

Like here
https://www.terraform.io/docs/providers/aws/r/spot_fleet_request.html#wait_for_fulfillment

t0mk:
@jasmingacic you can wait for fullfillment like in the spot market request test in packngo:
https://github.com/packethost/packngo/blob/master/spotmarketrequest_test.go#L107

t0mk:
@jasmingacic it would be never completely guaranteed, but it would be very close, if you'd pick a facility with lot of free capacity (ewr1), `baremetal_0`, high spot price (usual market price * 3), and low `devices_max` (3). HashiCorp is running the tests for every provider in their own Teamcity and they are used to failing acceptance tests randomly.

But you have a good point, and I think that the lack of guarantee can be fixed with timeouts
https://www.terraform.io/docs/configuration/resources.html#timeouts
example use: https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L26

I think there should be a Create and Delete timeouts.

I think it would be really useful for users if they could have the resource wait for the spot devices to be created, and for the spot devices to be removed. There is no way in the Packet Portal to check a status of the spot market request, so having the option to make TF to ensure it's done would be convenient IMO. So I propose 2 additional params:
- `wait_for_fullfilment`
- `wait_for_termination`

I won't force you to improve the test to include the timeout (although I think that if the test doesn't create devices and then checks that they are removed, it's useless). I will try to create and destroy a couple of fleet requests manually on my own and comment my experiences here.

jasmingacic:
@t0mk were you able to test this?

radeksimko:
Hi folks,
the PR looks generally reasonable, I'll leave the details on @t0mk , but I want to say that I agree with him that tests in this PR should really test the resource more thoroughly, i.e. we should check whether the spot instance was actually launched.

Implementing timeouts would be helpful not only for the purpose of testing, but also for users.

@jasmingacic Do you mind looking into this?

jasmingacic:
@t0mk I'm looking into it. I'm not sure how to implement such timeout when we are expecting more than one device to become active when the  max_bid_price is available an, because only then it will trigger device creation. Having wait feature in the resource could mean that TF might hang for a 24 hours or even longer. 
@radeksimko Do you have a suggestion how to solve this?

radeksimko:
> Having wait feature in the resource could mean that TF might hang for a 24 hours or even longer. 

We can set the timeout in test configs to something high, but still reasonable, e.g. 1 hour and we're totally ok with the test timing out from time to time. That's just (unfortunate) nature of many of our tests, not just Packet.

As @t0mk mentioned above we can reduce the chance of this happening by picking the right configuration, region etc.

jasmingacic:
@t0mk @radeksimko I've added wait functions for create and delete. Both will not wait for devices unless an optional parameter `wait_for_devices` is passed in.

t0mk:
@jasmingacic I sumbitted review of cosmetics. I will now run a couple of tests with TF templates.

packet/resource_packet_spot_market_request.go
```diff
@@ -0,0 +1,306 @@
...
+			},
+			"project_id": &schema.Schema{
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"wait_for_devices": &schema.Schema{
+				Type:     schema.TypeBool,
+				Optional: true,
+				ForceNew: true,
```
t0mk: You don't want to set ForceNew in `wait_for_devices`. If user creates a SMR, and then wants to remove it and doesn't want to wait, he might change it from False to true, and then TF will re-create the SMR.

And indeed tt's `ForceNew: false` in aws `spot_fleet_request`:
https://github.com/terraform-providers/terraform-provider-aws/blob/master/aws/resource_aws_spot_fleet_request.go#L47

t0mk:
I tested a bit with >1 devices, and with timeouts, I also run the acceptance test. And the resource seems to work.

t0mk:
@jasmingacic I improved the doc a bit. I tested yesterday and the resource seems to work. I have no objections anymore. Please check my last commit and let me know if you're fine with it and I will merge.

jasmingacic:
@t0mk go ahead and merge it

PR #74, opened 2018-09-26
fix broken documentation references

normanjoyner:
Updates `help.packet.net` documentation references to use new url structure


t0mk:
Thanks for the fix! I updated a couple of other links in the docs.

The new documentation will be visible after new release, which we might do after #72 is finished.

PR #83, opened 2018-12-17
Add resource for querying facilities

### LINKED ISSUE #32: failover if certain server is not available in particular facility at the moment
### Terraform Version
0.10.7

### Affected Resource(s)
- packet_device

### Terraform Configuration Files
```hcl
 provider "packet" {
   auth_token = ""
 }

 variable "project_id" {
   type = "string"
 }

 resource "packet_device" "first-deploy" {
   hostname         = "tf.test"
   plan             = "baremetal_0"
   facility         = "sjc1"
   operating_system = "centos_7"
   billing_cycle    = "hourly"
   project_id       = "${var.project_id}"
 }
```

### Debug Output
https://gist.github.com/kphatak/88cb22fdaea8f1b607868e75dfc46548

### Expected Behavior
Ideally, provider should accept multiple facilities and host types and retry on failure with a certain type of error

### Actual Behavior
Provider errors out with servers unavailable error.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. check what server type is unavailable from UI for a particular facility
2. Provision using terraform for that facility and server type

t0mk:
This PR adds a resource for querying facilities, in order to allow users to take advantage of the facility features from the API; and to select facility automatically.

This is a rework of dynamic "region" selection functionality. It addresses:
- a previous PR, where dynamic selection of facilities on device creation was refused by @radeksimko - https://github.com/terraform-providers/terraform-provider-packet/pull/76#pullrequestreview-179394967
- consultation with @tombuildsstuff in the TF-committers Slack channel

The new resource is a query much like https://www.terraform.io/docs/providers/random/r/string.html  rather than an item in the API. I decided not to do it as a datasource, because of the utilization level filter - utilization of datacenters will vary and the facility query should be executed only on argument change - that's why it offers the `keepers` attr, like the random\_ resources.

It closes #32 and #80.

### DIFF
--- a/packet/provider.go
+++ b/packet/provider.go
@@ -35,6 +35,7 @@ func Provider() terraform.ResourceProvider {
 			"packet_ip_attachment":       resourcePacketIPAttachment(),
 			"packet_spot_market_request": resourcePacketSpotMarketRequest(),
 			"packet_vlan":                resourcePacketVlan(),
+			"packet_facility":            resourcePacketFacility(),
 		},
 
 		ConfigureFunc: providerConfigure,
--- a/packet/resource_packet_facility.go
+++ b/packet/resource_packet_facility.go
@@ -1,0 +1,151 @@
+package packet
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/hashicorp/terraform/helper/validation"
+	"github.com/packethost/packngo"
+)
+
+func resourcePacketFacility() *schema.Resource {
+	return &schema.Resource{
+		Create: resourcePacketFacilityCreate,
+		Read:   resourcePacketFacilityRead,
+		Delete: schema.RemoveFromState,
+		Schema: map[string]*schema.Schema{
+			"keepers": {
+				Type:     schema.TypeMap,
+				Optional: true,
+				ForceNew: true,
+			},
+			"slugs": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+			},
+			"features": {
+				Type:     schema.TypeSet,
+				Elem:     &schema.Schema{Type: schema.TypeString},
+				ForceNew: true,
+				MinItems: 1,
+				Optional: true,
+			},
+			"plan": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ForceNew:     true,
+				ValidateFunc: validation.StringInSlice(packngo.DevicePlans, false),
+			},
+			"utilization": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringInSlice(packngo.UtilizationLevels, false),
+				ForceNew:     true,
+			},
+		},
+	}
+}
+
+func filterOnPlan(slugs []string, cr *packngo.CapacityReport, plan string) []string {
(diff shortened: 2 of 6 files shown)


t0mk:
@mbfrahry Thank You for looking at this PR!

I initially wanted to do this as a datasource, but it did not work. It's same reason why the `random` provider contains resources rather than datasources - I want to have the query executed only once. If this would be a datasource, I couldn't link it to a compute resource as in

```
locals {
    plan = "baremetal_0"
}

resource "packet_facility" "example_facility" {
  plan             = "${local.plan}"
  features         = ["global_ipv4"]
  utilization       = "normal"
  keepers       {
    when = "2018-12-15"
  }
}

resource "packet_device" "server" {
  hostname         = "tftest"
  plan             = "${local.plan}"
  facility         = "${packet_facility.example_facility.slugs.0}"
  operating_system = "ubuntu_16_04"
  billing_cycle    = "hourly"
  project_id       = "${packet_project.cool_project.id}"
}
```

... because the utilization can change, and then the link would ForceNew a new compute resource.

mbfrahry:
Hmmmm, I'm seeing your reasoning but I think we can do this a little differently. If `utilization` has the ability to change, we can just not set in the read function or remove the `ForceNew` tag from it to prevent the link destruction from triggering. Are either of those options? If not, then I'm fine with this being the way it gets shipped.

t0mk:
@mbfrahry thanks for your response! I have learned a bit of more terraform as of now, and I think it's possible to make random selection of available facility with `random_shuffle`, shielding out the keepers functionality. The facility selection could then work as a datasource.

I will get back you when I have the re-work ready.

t0mk:
@mbfrahry I still can't see another way to do facility selection in a way, that it would not trigger dependent resource (for example a device) re-creation when result of the faciltiy query changes.

I did try to rewrite this to a datasource, and then tried to shadow out the change in output of this datasource with random_shuffle like this:

```hcl
locals {
    plan = "baremetal_1e"
}

data "packet_facility" "test" {
    features = ["layer_2", "storage"]
    plan = "${local.plan}"
    utilization = "normal"
}

resource "random_shuffle" "test" {
    input = ["${sort(data.packet_facility.test.slugs)}"]
    result_count = 1
    keepers = {
        "when" = "2019-02-01"
    }
}

output "selected_facilties_accoring_to_criteria" {
    value = "${data.packet_facility.test.slugs}"
}

output "randomly_picked_facility" {
    value = "${random_shuffle.test.result}"
}

resource "packet_project" "test" {
    name = "test"
}

resource "packet_device" "web1" {
    hostname         = "tftest"
    plan             = "${local.plan}"
    facility         = "${random_shuffle.test.result.0}"
    operating_system = "ubuntu_16_04"
    billing_cycle    = "hourly"
    project_id       = "${packet_project.test.id}"
}
```

but even if I sort it, the list of the slugs from the facility data source will change when the utilization in Packet datacenters changes. And then, another facility might bubble through the random_shuflle, and terraform will try to re-created the dependent device. **If I make it a datasource, it will work, but it will be useless - people can't use it because it will be constantly causing dependent-resource re-creating.**

I think the only way to control the execution of the facility query is to contain the result in one resource, and re-trigger only when attibutes change (the filtering attrinbutes, or the keepers) - the current state of this PR.

t0mk:
This was solved by the `facilities` field in `packet_device`.

PR #86, opened 2018-12-18
Layer 2 support: Device ports and VLAN attachment

t0mk:
This PR goes towards Packet L2-features support in the provider. It adds ports listing to device resource and a new vlan_attachment resource.


dustnic:
When will this be merged?? :-)

t0mk:
hi @dustnic, thanks for your interest. This is not complete yet, the API port conversion between L2 and L3 is missing. Some of the L2 issues in the Packet API are not so straightforward to do in Terraform.

t0mk:
This is blocked by a bug in Packet API - inconsistent reporting of the device network_type:
https://github.com/packethost/packngo/issues/133

t0mk:
This is done. Please review.

packet/resource_packet_device.go
```diff
@@ -328,9 +364,53 @@ func resourcePacketDeviceCreate(d *schema.ResourceData, meta interface{}) error
...
+		_, err := client.DevicePorts.DeviceToNetworkType(newDevice.ID, targetNetworkState)
+		if err != nil {
+			return err
+		}
+	}
+
 	return resourcePacketDeviceRead(d, meta)
 }
 
+func getNetworkInfo(ips []*packngo.IPAddressAssignment) ([]map[string]interface{},
```
patrickdevivo: could it be worth having this func return a dedicated struct rather than so many values?

packet/resource_packet_port_vlan_attachment.go
```diff
@@ -0,0 +1,159 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"port_name": {
+				Type:     schema.TypeString,
+				Required: true,
+				ForceNew: true,
+			},
+			"vlan_id": {
```
johnstudarus: Any reason a VLAN-ID is a string and not an int?

packet/resource_packet_port_vlan_attachment.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return err
+	}
+	forceBond := d.Get("force_bond").(bool)
+	if forceBond && (len(portPtr.AttachedVirtualNetworks) == 0) {
+		_, _, err = client.DevicePorts.Bond(&packngo.BondRequest{PortID: pID, BulkEnable: false})
+		if err != nil {
+			return friendlyError(err)
+		}
+
```
johnstudarus: unneeded white page

packet/resource_packet_port_vlan_attachment.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
```
johnstudarus: unneeded white page

packet/resource_packet_port_vlan_attachment_test.go
```diff
@@ -0,0 +1,170 @@
...
+func testAccCheckPacketPortVlanAttachmentConfig_L2Individual(name string) string {
+	return fmt.Sprintf(`
+resource "packet_project" "test" {
+    name = "%s"
+}
+
+resource "packet_device" "test" {
+  hostname         = "test"
+  plan             = "m1.xlarge.x86"
+  facility         = "ewr1"
```
johnstudarus: just FYI - I've been told DFW2 has deeper pools of resources that EWR1. Or can this be set to a list of all sites that support layer2 using facilities rather than facility?

packet/resource_packet_port_vlan_attachment_test.go
```diff
@@ -0,0 +1,170 @@
...
+	"github.com/packethost/packngo"
+)
+
+func testAccCheckPacketPortVlanAttachmentConfig_L2Individual(name string) string {
+	return fmt.Sprintf(`
+resource "packet_project" "test" {
+    name = "%s"
+}
+
+resource "packet_device" "test" {
```
johnstudarus: This config file is duplicated in the code - perhaps use a static somewhere and just replace the network-type in each test function?

packet/resource_packet_device.go
t0mk: OK!

packet/resource_packet_device.go
```diff
@@ -328,9 +364,53 @@ func resourcePacketDeviceCreate(d *schema.ResourceData, meta interface{}) error
...
+		networks = append(networks, network)
+
+		// Initial device IPs are fixed and marked as "Management"
+		if ip.Management {
+			if ip.AddressFamily == 4 {
+				if ip.Public {
+					host = ip.Address
+					ipv4SubnetSize = ip.CIDR
+					public4 = ip.Address
+				} else {
```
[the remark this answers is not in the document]
t0mk: yes, they're necessary, they assign to the private4 and public6 vars, which are then returned.

I have gofmt run on every `:w` in vim, it doens't complain

packet/resource_packet_port_vlan_attachment.go
t0mk: @johnstudarus It's not the VXLAN, but the UUID of the Packet API resource. Do you think that the vxlan int would be more convenient to pass to the attachment? Also a question for @patrickdevivo

packet/resource_packet_port_vlan_attachment_test.go
t0mk: I agree there's duplication, but I think it's better that the whole config is in the same string literal - it's easier to understand the testing HCL if it's all in one place.
Furthermore this is testing code, so I think the duplication doesn't do too much harm.

packet/resource_packet_port_vlan_attachment_test.go
t0mk: thanks for the tip

PR #89, opened 2019-01-16
keep device networks in specific order

t0mk:
This PR sorts the network connections in device's `network` list to specific order - first is private IPv4, second is public IPv4 and third is IPv6. This will allow to access attributes of a particular interface configuration, which is necessary in BGP setup in #88.


t0mk:
test pass, I merge

PR #92, opened 2019-02-04
Hotfix of network attribute order for packet_device

t0mk:
I have changed the order of the elements in the network List attribute in #89. It was not a great idea considering how many users depend on the implicit order of the network configuration:

https://github.com/search?p=2&q=%22network.0.address%22+ssh+packet+terraform&type=Code

This PR reverts the order of the items in the network, so that it's like before, i.e.

0. Public IPv4
1. IPv6
2. Private IPv4

plus documentation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
-## 1.4.0 (Unreleased)
+## 1.3.1 (Unreleased)
+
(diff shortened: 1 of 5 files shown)


mmlb:
:+1:

pereztr5:
👍

PR #96, opened 2019-02-07
Project ssh keys

t0mk:
This PR adds a resource for project SSH keys, keys which are (not user-, but) project-specific.

If project SSH keys are passed to a new device, no user keys are injected.

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"path"
+	"path/filepath"
 	"reflect"
 	"regexp"
 	"sort"
@@ -86,7 +87,6 @@ func resourcePacketDevice() *schema.Resource {
 				Required: true,
 				ForceNew: true,
 			},
-
 			"state": {
 				Type:     schema.TypeString,
 				Computed: true,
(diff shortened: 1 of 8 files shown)


packet/resource_packet_project_ssh_key_test.go
```diff
@@ -0,0 +1,86 @@
...
+    project_id = "${packet_project.test.id}"
+}
+
+resource "packet_device" "test" {
+    hostname            = "test"
+    plan                = "baremetal_0"
+    facility            = "ewr1"
+    operating_system    = "ubuntu_16_04"
+    billing_cycle       = "hourly"
+	project_ssh_key_ids = ["${packet_project_ssh_key.test.id}"]
```
patrickdevivo: extra tab here?

packet/resource_packet_device.go
```diff
@@ -280,13 +291,14 @@ func resourcePacketDeviceCreate(d *schema.ResourceData, meta interface{}) error
...
 
+	projectKeys := d.Get("project_ssh_key_ids.#").(int)
+	if projectKeys > 0 {
+		createRequest.ProjectSSHKeys = convertStringArr(d.Get("project_ssh_key_ids").([]interface{}))
+	}
+
 	tags := d.Get("tags.#").(int)
 	if tags > 0 {
-		createRequest.Tags = make([]string, 0, tags)
-		for i := 0; i < tags; i++ {
```
johnstudarus: would this be better with an iterator?

website/docs/r/project_ssh_key.html.markdown
```diff
@@ -0,0 +1,57 @@
...
+layout: "packet"
+page_title: "Packet: packet_project_ssh_key"
+sidebar_current: "docs-packet-resource-project-ssh-key"
+description: |-
+  Provides a Packet Project SSH key resource.
+---
+
+# packet_project_ssh_key
+
+Provides a Packet project SSH key resource to manage project-specific SSH keys. On contrary to user SSHK keys, project SSH keys are used to exclusively populate `authorized_keys` in new devices.
```
johnstudarus: ^SSHK^SSH

packet/resource_packet_device.go
t0mk: This is removed code. The for loop is moved to the convertStringArr function,

PR #99, opened 2019-02-12
Global floating IP block reservation

t0mk:
this PR extends the IP reservation resource to allow global IPv4

### DIFF
--- a/packet/datasource_packet_precreated_ip_block.go
+++ b/packet/datasource_packet_precreated_ip_block.go
@@ -59,6 +59,10 @@ func dataSourcePacketReservedIPBlockRead(d *schema.ResourceData, meta interface{
 	public := d.Get("public").(bool)
 	global := d.Get("global").(bool)
 
+	if !public && global {
+		return fmt.Errorf("Private (non-public) global IP address blocks are not supported in Packet")
+	}
+
 	fval, fok := d.GetOk("facility")
 	if fok && global {
 		return fmt.Errorf("You can't specify facility for global IP block - addresses from global blocks can be assigned to devices across several facilities")
--- a/packet/datasource_packet_precreated_ip_block.go
+++ b/packet/datasource_packet_precreated_ip_block.go
@@ -9,62 +9,41 @@ import (
 )
 
 func dataSourcePacketPreCreatedIPBlock() *schema.Resource {
+	s := packetIPComputedFields()
+	s["project_id"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Required: true,
+	}
+	s["global"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Optional: true,
+		Default:  false,
+	}
+	s["public"] = &schema.Schema{
+		Type:     schema.TypeBool,
+		Required: true,
+	}
+
+	s["facility"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Optional: true,
+	}
+	s["address_family"] = &schema.Schema{
+		Type:     schema.TypeInt,
+		Required: true,
+	}
+	s["cidr_notation"] = &schema.Schema{
+		Type:     schema.TypeString,
+		Computed: true,
+	}
+	s["quantity"] = &schema.Schema{
+		Type:     schema.TypeInt,
+		Computed: true,
+	}
+
 	return &schema.Resource{
-		Read: dataSourcePacketReservedIPBlockRead,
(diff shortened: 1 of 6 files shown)


t0mk:
This is ready. I checked that IP address blocks created with previous provider version will not generate diff.

packet/datasource_packet_precreated_ip_block.go
```diff
@@ -78,13 +57,32 @@ func dataSourcePacketReservedIPBlockRead(d *schema.ResourceData, meta interface{
 	}
 	ipv := d.Get("address_family").(int)
 	public := d.Get("public").(bool)
-	facility := d.Get("facility").(string)
+	global := d.Get("global").(bool)
```
johnstudarus: Can something be a global but not public? Is that a worthwhile check?

packet/datasource_packet_precreated_ip_block.go
```diff
@@ -78,13 +57,32 @@ func dataSourcePacketReservedIPBlockRead(d *schema.ResourceData, meta interface{
...
+
+	fval, fok := d.GetOk("facility")
+	if fok && global {
+		return fmt.Errorf("You can't specify facility for global IP block - addresses from global blocks can be assigned to devices in any facility")
```
johnstudarus: "...assigned to devices across several facilities. For a single facility please use a local IP block."

Better explains that global IPs are to be used by many devices across may facilities.

johnstudarus: I'd move the "if fok && global..." check into the "if fok" block below for better reading flow.

packet/resource_packet_reserved_ip_block.go
```diff
@@ -101,33 +131,58 @@ func resourcePacketReservedIPBlockCreate(d *schema.ResourceData, meta interface{
...
+		return err
+	}
+	d.Set("type", typ)
 	d.Set("public", reservedBlock.Public)
 	d.Set("management", reservedBlock.Management)
 	d.Set("manageable", reservedBlock.Manageable)
 	if reservedBlock.AddressFamily == 4 {
 		d.Set("quantity", ipv4CIDRToQuantity[reservedBlock.CIDR])
 	} else {
 		// In Packet, reserved IPv6 block is allocated when device is run in a proejct.
```
johnstudarus: ^proejct^project

website/docs/r/reserved_ip_block.html.markdown
```diff
@@ -30,22 +31,32 @@ resource "packet_reserved_ip_block" "two_elastic_addresses" {
...
+* `facility` - (Optional) Facility where to allocate the public IP address block, makes sense only for type==public_ipv4
+* `type` - (Optional) Either "global_ipv4" or "public_ipv4", defaults to "public_ipv4" for backward compatibility
+
 
 ## Attributes Reference
 
 The following attributes are exported:
 
-* `facility` - The facility where the addresses are
+* `facility` - The facility where the block was allocated, empty for global blocks
```
johnstudarus: ^empty^must be empty

packet/datasource_packet_precreated_ip_block.go
t0mk: Thanks, I like your error message better.

If I'd move it to the `if fok {` block, I'd need to check the `global` flag (and thus assert the same conditoin) twice - secondly in the `} else {' block for `!global`.
This way it's checked once, deduping and making the assert quite explicit, I think.

packet/datasource_packet_precreated_ip_block.go
t0mk: At the moment there can't be any IP blocks which are both global and not-public. A "global" VPN would be a case for non-public global block. I will add the check.

website/docs/r/reserved_ip_block.html.markdown
t0mk: I agree that the facility-global_ipv4 constraint should be written explicitly. I added it to the "Argument Reference".

In the "Attributes Reference" section, I think it's proper to only _declare_ that the facility _is_ empty for global_ipv4.

t0mk:
@patrickdevivo @johnstudarus thanks for review. I run the acceptance tests and some are flaking, I think it's because something is wrong inside the Packet API. I will try tomorrow and if it goes through, I will merge.

t0mk:
I removed the commit disabling SpotMarketRequest test, now running the suite again.

t0mk:
the tests pass now, merging

PR #100, opened 2019-02-14
[doc] Use real plans in the doc

Miouge1:
I think that giving real example will more, AFAIK `plan = "baremetal_1"` is not something that would work.


t0mk:
@Miouge1 thanks. I updated all the other occurences of `baremetal_*` in this repo.

t0mk:
I rebased this on current master with the Go modules.

PR #102, opened 2019-02-15
vendor: github.com/packethost/packngo@caf7bb8

radeksimko:
This is to bump Packet SDK to a go-mod-aware version.

PR is a result of the following:

```
go get github.com/packethost/packngo@caf7bb8
go mod tidy
go mod vendor
```


t0mk:
Acceptance tests passed, looks good :+1: Go ahead and merge.

Issue #66, opened 2018-04-27
Improved error messages 


vielmetti:
I'm trying to sort out what is going on with the CNCF cross-cloud CI system for Kubernetes, which uses Terraform to set up resources. (http://cncf.ci) As a piece of that I'm looking through logs and come up blank because some of the error messages are not as good as they might be.  See e.g. from https://gitlab.cncf.ci/cncf/cross-cloud/-/jobs/81265

```
Error: Error applying plan:

2 error(s) occurred:
* module.master.packet_device.masters[1]: 1 error(s) occurred:
* packet_device.masters.1: 
* module.master.packet_device.masters[0]: 1 error(s) occurred:
* packet_device.masters.0: 
```

vielmetti:
Bringing this to the attention of @t0mk - the question I have is whether the Packet API is correctly throwing an error or not, or if Terraform is silently discarding it.

t0mk:
you cna try to set TF_LOG=DEBUG env var for the terraform runs (especially for `terraform apply`) to see more of what's going on.

Or you could investigate from the Packet API side, and see what's happens there on the tf run.

Otherwise, it's impossible to say what's going on without the HCL files - that's why they are mentioned in the Issue Template for every TF provider.

vielmetti:
Thanks @t0mk . All of the provisioning parts for this system are in the https://github.com/crosscloudci/cross-cloud repo, especially https://github.com/crosscloudci/cross-cloud/tree/master/packet

t0mk:
@BRMatt thanks, for the submission.

I just tried to create device with underscore in the name, so that I get API error, and the `terraform apply` will show:

```
Error: Error applying plan:

1 error(s) occurred:

* packet_device.web1: 1 error(s) occurred:

* packet_device.web1: Hostname is not a valid hostname (can only contain alphanumeric values, dashes, and periods)
```

Do you have any idea what the error could be, so that I can replicate and debug this? I don't suggest you use TF_LOG=DEBUG all the time, maybe just once, to see what is causing the mute message.

jasmingacic:
@vielmetti This one is really hard to replicated. I noticed that the code is capturing all possible errors that might come from the API. So I believe that in some occasions API doesn't return any error messages.

Should we close this one?

vielmetti:
Hi @jasmingacic - I'm OK with closing this out for now, with the notation that if the problem reappears we'll aim to chase it from the Packet API side and not the Terraform side.

Issue #24, opened 2017-09-09
Provide a useful error message if auth_token is invalid


bzub:
I noticed that if `auth_token` is set, but invalid, we do not seem to output a useful error message.

The only case I've tested so far is creating a `packet_ssh_key` resource and `api_token` set to `""` empty string. I'll test to see if this is for all packet resources, and show an example in an update to this issue.

In provider.go we do not do any validation against `auth_token` currently. I assume the packet API returns an unauthorized response, which we should also let the user know of if anything gets past validation.

t0mk:
I confirm this, I run simple acceptance test with wrong PACKET_AUTH_TOKEN, and output is not very helpful:

```
--- FAIL: TestAccPacketReservedIPBlock_Basic (0.60s)
	testing.go:449: Step 0 error: Error applying: 1 error(s) occurred:
		
		* packet_project.foobar: 1 error(s) occurred:
		
		* packet_project.foobar: 
FAIL
exit status 1

```

PR #104, opened 2019-02-16
create useful error string when none generated from packngo

johnstudarus:
in specific, when an invalid token is used, a 401 will be generated without an error message
this now passes back up the 401 and an error string to the user rather than a blank message

previously, the user would be presented with no error string and would have to turn on TF_LOG=DEBUG
to figure out what happened

### DIFF
--- a/packet/errors.go
+++ b/packet/errors.go
@@ -2,6 +2,7 @@ package packet
 
 import (
 	"net/http"
+	"strconv"
 	"strings"
 
 	"github.com/packethost/packngo"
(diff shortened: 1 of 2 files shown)


johnstudarus:
This resolves: https://github.com/terraform-providers/terraform-provider-packet/issues/66

johnstudarus:
This resolves: https://github.com/terraform-providers/terraform-provider-packet/issues/24

t0mk:
@johnstudarus thanks, this looks good to me.

I've run acceptances tests in Hashicorp TeamCity, and if they all pass, I merge.

I will need to do something similar in the isNotFound err - at the moment it doesn't work.

johnstudarus:
When is the isNotFound error called? When importing a previous asset? I'd be happy to test it out and fix as appropriate.

t0mk:
@johnstudarus you can see the `isNotFound` on some places in the provider code. it's called when you create a resource with Terraform, then remove it out of TF (maybe via portal), and then call terraform again (not sure if `terraform plan` is enough, maybe `refresh` or even `apply`). 

The general provider policy is: In the <Resource>Read function, if the resource is not found by ID, the Read function should SetID("") and return nil.

The challenge is to recognize that the error is HTTP 404, and not some other type.

To test it, you can 
1. `terraform apply` on simple template with project and small device. 
2. remove the device in the web portal
3. run `terraform plan` or `refresh`

johnstudarus:
I found three places where it is not checking isNotFound...

packet/resource_packet_reserved_ip_block.go
packet/resource_packet_spot_market_request.go
packet/resource_packet_volume_attachment.go

All the other Reads are checking isNotFound ok...

I'm moving this discussion to the open issue at:

https://github.com/terraform-providers/terraform-provider-packet/issues/108

johnstudarus:
PR submitted to resolve the isNotFound issue.

https://github.com/terraform-providers/terraform-provider-packet/pull/111

PR #105, opened 2019-02-17
clarify the payment_method_id in project.html.markdown

larrymapacket:
the 'payment_method_id' needs to belong to the same 'organization_id'

### DIFF
--- a/website/docs/r/project.html.markdown
+++ b/website/docs/r/project.html.markdown
@@ -8,13 +8,13 @@ description: |-
 
 # packet\_project
 
-Provides a Packet Project resource to allow you manage devices
+Provides a Packet project resource to allow you manage devices
 in your projects.
 
 ## Example Usage
 
 ```hcl
-# Create a new Project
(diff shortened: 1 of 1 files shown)


website/docs/r/project.html.markdown
```diff
@@ -55,7 +55,7 @@ The `bgp_config` block supports:
 The following attributes are exported:
 
 * `id` - The unique ID of the project
-* `payment_method_id` - The UUID of payment method for this project.
+* `payment_method_id` - The UUID of payment method for this project,needing to belong to the same organization_id.
```
johnstudarus: ^, needing^ which needs

t0mk: @larrymapacket, I would reword to
"The UUID of payment method for this project, which needs to belong to the same organization (default, or the one passed with organizaition_id)."

t0mk:
@larrymapacket Thank you for the PR!

PR #106, opened 2019-02-17
[docs] how to use the dev plugin in TF

johnstudarus:
instructions on how to set up TF to use the newly developed plugin within a development environment


t0mk:
thanks @johnstudarus, this is useful doc

Issue #108, opened 2019-02-18
isNotFound doesn't work


johnstudarus:
Per @t0mk , the isNotFound error routine doesn't work. Just a place holder until he can elaborate or someone else tries it out.

johnstudarus:
isNotFound on some places in the provider code. it's called when you create a resource with Terraform, then remove it out of TF (maybe via portal), and then call terraform again (not sure if terraform plan is enough, maybe refresh or even apply).

The general provider policy is: In the Read function, if the resource is not found by ID, the Read function should SetID("") and return nil.

The challenge is to recognize that the error is HTTP 404, and not some other type.

To test it, you can

terraform apply on simple template with project and small device.
remove the device in the web portal
run terraform plan or refresh

johnstudarus:
The following four Read instances are not checking isNotFound. This code needs to be modified with the isNotFound checking...

packet/resource_packet_reserved_ip_block.go
packet/resource_packet_spot_market_request.go
packet/resource_packet_volume_attachment.go

PR #111, opened 2019-02-19
Handle resources that have been removed outside of TF

johnstudarus:
aligning error handling to be the same as the rest of the code base
properly handle when a resource has been removed outside of Terraform

resolves https://github.com/terraform-providers/terraform-provider-packet/issues/108

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ IMPROVEMENTS:
 BUG FIXES:
 
 - [#104](https://github.com/terraform-providers/terraform-provider-packet/pull/104) Fix empty error messages on invalid credentials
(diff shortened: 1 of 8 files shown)


t0mk:
thanks @johnstudarus, the checks that you added are needed.

The problem is though, that the error that we check with `isNotFound` must be first processed with the `friendlyError` fucntion, otherwise the StatusCode is not present, and the `isNotFound` logic doesn't kick in.

I will add friendlyError to the isNotFound function, and remote unnecessary calls of the friendlyError from the Read functions.

t0mk:
@johnstudarus please check my changes, and give me green light.

Don't worry about the conflict with master, I will rebase it before merge.

johnstudarus:
lgtm

PR #112, opened 2019-02-21
Allow facilities which are not yet in packngo

smintz:


t0mk:
hi @smintz, thanks for your PR, but I don't think this is good idea. Doing API request on validation is unprecedented in other providers (I took a look in aws and google). It will slow down `plan` and `apply`, and it introduces point of failure without error handling.

If more people would complain that the validation of the facility is too strict, I would rather remove the validation check, and let the API to err; rather than doing an API request for each validatoin.

I've just checked and the API reports 17 facilities, and there's 17 facilties in packngo. Also, we know that if there's a new facility added, it takes ~1 business  day to get it through to the provider and re-release - we had this happening in the beinning of this month:
https://github.com/terraform-providers/terraform-provider-packet/blob/master/CHANGELOG.md#132-february-06-2019

What exactly is your problem with the facilities?

smintz:
@t0mk: The call to api is only a fallback in case the facility is not in the list of `packngo.Facilities`. In most cases it will never call the API.

The API can expose private facilities per the token provided. And we also needs this for packet internal use when installing new sites.

t0mk:
@smintz I see. What about removing the check then? I am really strongly against API request in the validation, even as a fallback - typos will issue the API request, `bad facility` will be reported on wrong credentials, etc.

We'd only need to ensure that "any" is not allowed in the `facility` legacy argument in the device resource, because that breaks the Teraform state. "any" is still allowed in davice::`faclities`

smintz:
Sounds good.

smintz:
@t0mk: Updated.

t0mk:
@smintz Thansk. I launched the acceptance tests in Hashicorp CI. If they pass I merge.

Do you need this released now, or can it wait till next release? ~2 weeks,

t0mk:
@smintz, it's the `v interface{}` arg of the validation function that holds the value. The `k` is the key.

I fixed the validation and added a test.

t0mk:
.. otherwise the PR looks good. The acceptance tests passed, except for some API error in the Organization creating cases, but that's probably unreleated to this change.

philipcristiano:
I think a release sooner would help us out quite a bit, we are trying to release some new facilities and 2 weeks would slow us down.

t0mk:
@philipcristiano OK, I asked for a minor release (1.4.1), should be out later today.

philipcristiano:
Thank you for the quick assistance! We're using the new build and it's working out for us.

PR #116, opened 2019-03-06
Fix unordered lists

blalor:
In Markdown unordered lists need an empty line before their start.


t0mk:
thanks @blalor I will merge this and the docs will be updated on next release.

PR #144, opened 2019-04-17
Add support for Native VLAN

t0mk:
This PR makes it possible to mark attached VLAN as "native" to a port.


t0mk:
please review @patrickdevivo

PR #148, opened 2019-05-07
bump packngo to support IP address type specs when creating device

t0mk:


t0mk:
Tests pass, except for one flake. I run the TestAccPacketPortVlanAttachment_L2Individual test locally and it passed. I merge.

PR #151, opened 2019-05-17
docs/device: Document ports attribute

radeksimko:


t0mk:
@radeksimko thanks for the upkeep. Looks good!

PR #152, opened 2019-05-23
resource/packet_port_vlan_attachment: Avoid parallel assignment

radeksimko:
Packet's API reports HTTP 500 Internal Server Error when 2+ requests are issued for assigning a VLAN to a port in parallel, as demonstrated by attached test suite.

I have already contacted Packet support and they said the API doesn't support parallel assignment at this point and they agreed the API should respond 4xx instead of 5xx in such cases. They plan to fix the API in that sense, but implementation/allowing parallel execution may take more time.

I'm therefore adding mutex to avoid parallel VLAN assignments until this is actually fixed/implemented.

### Before

```
$ make testacc TEST=./packet TESTARGS='-run=TestAccPacketPortVlanAttachment_HybridMultipleVlans'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./packet -v -run=TestAccPacketPortVlanAttachment_HybridMultipleVlans -timeout 120m
=== RUN   TestAccPacketPortVlanAttachment_HybridMultipleVlans
--- FAIL: TestAccPacketPortVlanAttachment_HybridMultipleVlans (180.58s)
    testing.go:568: Step 0 error: errors during apply:

        Error: POST https://api.packet.net/ports/4746fa57-fded-46ad-8cf1-b4c0f34affc8/assign: 500 Oh snap, something went wrong! We've logged the error and will take a look - please reach out to us if you continue having trouble.

          on /var/folders/31/hkndp1hj17s_kp632xk19j1h0000gn/T/tf-test887383766/main.tf line 23:
          (source code not available)



        Error: POST https://api.packet.net/ports/4746fa57-fded-46ad-8cf1-b4c0f34affc8/assign: 500 Oh snap, something went wrong! We've logged the error and will take a look - please reach out to us if you continue having trouble.

          on /var/folders/31/hkndp1hj17s_kp632xk19j1h0000gn/T/tf-test887383766/main.tf line 23:
          (source code not available)


FAIL
FAIL	github.com/terraform-providers/terraform-provider-packet/packet	180.611s
make: *** [testacc] Error 1
```

### After

```
$ make testacc TEST=./packet TESTARGS='-run=TestAccPacketPortVlanAttachment_HybridMultipleVlans'
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test ./packet -v -run=TestAccPacketPortVlanAttachment_HybridMultipleVlans -timeout 120m
=== RUN   TestAccPacketPortVlanAttachment_HybridMultipleVlans
--- PASS: TestAccPacketPortVlanAttachment_HybridMultipleVlans (292.21s)
PASS
ok  	github.com/terraform-providers/terraform-provider-packet/packet	292.240s
```

### DIFF
--- a/packet/resource_packet_port_vlan_attachment.go
+++ b/packet/resource_packet_port_vlan_attachment.go
@@ -63,7 +63,9 @@ func resourcePacketPortVlanAttachmentCreate(d *schema.ResourceData, meta interfa
 	pName := d.Get("port_name").(string)
 	vlanVNID := d.Get("vlan_vnid").(int)
 
-	dev, _, err := client.Devices.Get(deviceID, &packngo.GetOptions{Includes: []string{"virtual_networks,project,native_virtual_network"}})
+	dev, _, err := client.Devices.Get(deviceID, &packngo.GetOptions{
+		Includes: []string{"virtual_networks,project,native_virtual_network"},
+	})
 	if err != nil {
 		return err
 	}
@@ -89,6 +91,7 @@ func resourcePacketPortVlanAttachmentCreate(d *schema.ResourceData, meta interfa
 	if !portFound {
 		return fmt.Errorf("Device %s doesn't have port %s", deviceID, pName)
 	}
+
 	par := &packngo.PortAssignRequest{PortID: port.ID}
 	if vlanFound {
 		log.Printf("Port %s already has VLAN %d assigned", pName, vlanVNID)
@@ -109,6 +112,12 @@ func resourcePacketPortVlanAttachmentCreate(d *schema.ResourceData, meta interfa
 		}
 
 		par.VirtualNetworkID = vlanID
+
+		// Packet doesn't allow multiple VLANs to be assigned
+		// to the same port at the same time
+		lockId := "vlan-attachment-" + port.ID
+		packetMutexKV.Lock(lockId)
+		defer packetMutexKV.Unlock(lockId)
 
 		_, _, err = client.DevicePorts.Assign(par)
 		if err != nil {
--- a/packet/provider.go
+++ b/packet/provider.go
@@ -3,11 +3,13 @@ package packet
 import (
 	"time"
 
+	"github.com/hashicorp/terraform/helper/mutexkv"
 	"github.com/hashicorp/terraform/helper/schema"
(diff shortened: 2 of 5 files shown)


t0mk:
@radeksimko thanks for the PR, I vaguely remember seeing something like this in the API. 

Travis CI is faiiking, maybe the "github.com/hashicorp/terraform/helper/mutexkv" is missing in go,mod?
 
I will run the test and take better look tomorrow.

radeksimko:
@t0mk Yes, sorry - I forgot to run `go mod vendor`. It should be fixed now.

packet/resource_packet_port_vlan_attachment.go
```diff
@@ -110,13 +114,16 @@ func resourcePacketPortVlanAttachmentCreate(d *schema.ResourceData, meta interfa
 
 		par.VirtualNetworkID = vlanID
 
+		lockId := "vlan-attachment-" + id
```
t0mk: I'm not sure if the lockId is correct here. The `id` is combination of port and vlan IDs, and such combination should be unique, unless there's something wrong with the setup logic.

If the mutex is supposed to guard the VLAN, shouldn't it be
`lockId := "vlan-attachment-" + vlanID`
.. or similarly if it's supposed to guard the port

t0mk:
@radeksimko the tests run, i have just 1 comment on the lock Id,

packet/resource_packet_port_vlan_attachment.go
t0mk: I'm not sure if the lockId is correct here. The id is combination of port and vlan IDs, and such combination should be unique, unless there's something wrong with the setup logic.

If the mutex is supposed to guard the VLAN, shouldn't it be
lockId := "vlan-attachment-" + vlanID
.. or similarly if it's supposed to guard the port

radeksimko: Good point! Thanks for noticing. I ran a few tests to see if we should lock on port ID or VLAN ID and landed on the conclusion that it should be port ID (at least tests passed when I used port ID as a mutex key).

t0mk:
@radeksimko OK, Looks good, go ahead and merge!

PR #156, opened 2019-06-06
Rework operating systems datasource filtering and add tests

### LINKED ISSUE #155: data/packet_operating_system: Incorrect error handling
### Terraform Version

```
Terraform v0.12.0
+ provider.packet v2.2.1
```

### Terraform Configuration Files
```hcl
data "packet_operating_system" "esxi" {
  name             = "VMware ESXi"
  distro           = "vmware"
  version          = "BLABLAH"
  provisionable_on = "c1.xlarge.x86"
}
```

### Expected Behavior

```
data.packet_operating_system.esxi: Refreshing state...

Error: There are no operating systems that match the search criteria

  on main.tf line 1, in data "packet_operating_system" "esxi":
   1: data "packet_operating_system" "esxi" {

```

### Actual Behavior

```
data.packet_operating_system.esxi: Refreshing state...

Error: There is more than one operating system that matches the search criteria

  on main.tf line 1, in data "packet_operating_system" "esxi":
   1: data "packet_operating_system" "esxi" {

```

### Steps to Reproduce

1. `terraform apply`

---

I'm not sure if this is an API bug or provider bug - I haven't done any debugging at all, but it looked confusing enough for me to at least file it as an issue.

I assume it would be helpful if the "more than one" error message also contained more details about what was returned from the API - e.g. how many systems and which ones; with some reasonable limits obviously - we don't want to return all available OS in case of too loose match. I reckon <= 5-10 would suffice for some basic debugging?

t0mk:
This PR reworks the OS datasource filtering and add tests for the case reported in #155.

Fixes #155

### DIFF
--- a/packet/datasource_packet_operating_system.go
+++ b/packet/datasource_packet_operating_system.go
@@ -58,73 +58,60 @@ func dataSourcePacketOperatingSystemRead(d *schema.ResourceData, meta interface{
 		return err
 	}
 
-	final := []packngo.OS{}
-	temp := []packngo.OS{}
-
 	if nameOK {
+		temp := []packngo.OS{}
 		for _, os := range oss {
 			if strings.Contains(strings.ToLower(os.Name), strings.ToLower(name.(string))) {
 				temp = append(temp, os)
 			}
-			final = temp
 		}
+		oss = temp
 	}
 
-	if distroOK {
-		temp = []packngo.OS{}
-		if len(temp) == 0 {
-			final = oss
-		}
-		for _, v := range final {
+	if distroOK && (len(oss) != 0) {
+		temp := []packngo.OS{}
+		for _, v := range oss {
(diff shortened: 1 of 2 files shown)


packet/datasource_packet_operating_system.go
```diff
@@ -58,73 +58,60 @@ func dataSourcePacketOperatingSystemRead(d *schema.ResourceData, meta interface{
...
 				if po == provisionableOn.(string) {
 					temp = append(temp, v)
 				}
 			}
 		}
-		final = temp
+		oss = temp
 	}
-	log.Println("[DEBUG] RESULTS:", final)
+	log.Println("[DEBUG] RESULTS:", oss)
```
radeksimko: Nitpick: It would be nice to make this log message a little bit more descriptive.

PR #158, opened 2019-06-30
Add datasource for packet_device

t0mk:
This PR adds a datasource for packet_device

### DIFF
--- a/packet/datasource_packet_device.go
+++ b/packet/datasource_packet_device.go
@@ -60,7 +60,6 @@ func dataSourcePacketDevice() *schema.Resource {
 				Type:     schema.TypeString,
 				Computed: true,
 			},
-
 			"access_private_ipv4": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -95,18 +94,15 @@ func dataSourcePacketDevice() *schema.Resource {
 				},
 				Computed: true,
 			},
-
 			"root_password": {
 				Type:      schema.TypeString,
 				Computed:  true,
 				Sensitive: true,
 			},
-
 			"always_pxe": {
 				Type:     schema.TypeBool,
 				Computed: true,
 			},
-
 			"ipxe_script_url": {
 				Type:     schema.TypeString,
 				Computed: true,
@@ -120,22 +116,18 @@ func dataSourcePacketDevice() *schema.Resource {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
-
 						"gateway": {
 							Type:     schema.TypeString,
 							Computed: true,
 						},
-
 						"family": {
 							Type:     schema.TypeInt,
 							Computed: true,
 						},
-
 						"cidr": {
 							Type:     schema.TypeInt,
 							Computed: true,
 						},
-
 						"public": {
 							Type:     schema.TypeBool,
 							Computed: true,
--- a/packet/datasource_packet_device.go
+++ b/packet/datasource_packet_device.go
@@ -1,0 +1,65 @@
+package packet
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/terraform/helper/schema"
+	"github.com/packethost/packngo"
+)
+
+func dataSourcePacketDevice() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePacketDeviceRead,
+		Schema: map[string]*schema.Schema{
+			"hostname": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"project_id": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*packngo.Client)
+
+	hostname := d.Get("hostname").(string)
+	projectId := d.Get("project_id").(string)
+
+	ds, _, err := client.Devices.List(projectId, nil)
+	if err != nil {
+		return err
+	}
+
+	dev, err := findDeviceByHostname(ds, hostname)
+
+	if err != nil {
+		return err
+	}
+
+	d.SetId(dev.ID)
+	return nil
+}
+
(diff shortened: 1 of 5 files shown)


packet/datasource_packet_device.go
```diff
@@ -0,0 +1,259 @@
...
+			"network": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"address": {
+							Type:     schema.TypeString,
+							Computed: true,
+						},
+
```
mmlb: you should drop the empty-lines here or do the same in other `maps`

packet/datasource_packet_device.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/packethost/packngo"
+)
+
+func dataSourcePacketDevice() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePacketDeviceRead,
+		Schema: map[string]*schema.Schema{
+			"hostname": {
+				Type:     schema.TypeString,
+				Required: true,
```
mmlb: hostnames are not required to be unique vs id. It seems like the lookup should be with id and hostname be Computed to me.

packet/datasource_packet_device.go
```diff
@@ -0,0 +1,259 @@
...
+			results = append(results, d)
+		}
+	}
+	if len(results) == 1 {
+		return &results[0], nil
+	}
+	if len(results) == 0 {
+		return nil, fmt.Errorf("no device found with hostname %s", hostname)
+	}
+	return nil, fmt.Errorf("too many devices found with hostname %s (found %d, expected 1)", hostname, len(results))
```
mmlb: this seems like a pretty arbitrary limitation right? Why not look up via id?

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+
+* `hostname` - (Required) The device name
+* `project_id` - (Required) The id of the project in which the devices exists
+
+## Attributes Reference
+
+The following attributes are exported:
+
+* `access_private_ipv4` - The ipv4 private IP assigned to the device
+* `access_public_ipv4` - The ipv4 maintenance IP assigned to the device
```
mmlb: packet refers to these as `management` not `maintenance` is this a typo or tf convention?

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+* `hardware_reservation_id` - The id of hardware reservation which this device occupies
+* `id` - The ID of the device
+* `network` - The device's private and public IP (v4 and v6) network details. When a device is run without any special network configuration, it will have 3 networks: 
+  * Public IPv4 at `packet_device.name.network.0`
+  * IPv6 at `packet_device.name.network.1`
+  * Private IPv4 at `packet_device.name.network.2`
+  Elastic addresses then stack by type - an assigned public IPv4 will go after the management public IPv4 (to index 1), and will then shift the indices of the IPv6 and private IPv4. Assigned private IPv4 will go after the management private IPv4 (to the end of the network list).
+  The fields of the network attributes are:
+  * `address` - IPv4 or IPv6 address string
+  * `cidr` - bit length of the network mask of the address
```
mmlb: Should be `Bit`

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+  * `gateway` - address of router
```
mmlb: `Address`

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+  * `public` - whether the address is routable from the Internet
```
mmlb: `Whether`

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+* `network_type` - L2 network type of the device, one of "layer3", "layer2-bonded", "layer2-individual", "hybrid"
+* `operating_system` - The operating system running on the device
+* `plan` - The hardware config of the device
+* `ports` - Ports assigned to the device
+  * `name` - Name of the port (e.g. `eth0`, or `bond0`)
+  * `id` - ID of the port
+  * `type` - Type of the port (e.g. `NetworkPort` or `NetworkBondPort`)
+  * `mac` - MAC address assigned to the port
+  * `bonded` - Whether this port is part of a bond in bonded network setup
+* `root_password` - Root password to the server (disabled after 24 hours)
```
mmlb: The "disabled after 24 hours" seems worded in a confusing manner. Who disables it? ... Maybe just have `(if still available)" instead? You don't go into too much detail for other things (like `network_type`), probably shouldn't do so for the root password either.

website/docs/d/device.html.markdown
```diff
@@ -0,0 +1,76 @@
...
+* `ssh_key_ids` - List of IDs of SSH keys deployed in the device, can be both user and project SSH keys
```
mmlb: s/and/or/ ?

packet/datasource_packet_device.go
t0mk: thanks!

packet/datasource_packet_device.go
t0mk: the datasources are used to lookup compute resources by name. If you have ID, you can do terraform import.
https://www.terraform.io/docs/providers/google/d/datasource_compute_instance.html
https://www.terraform.io/docs/providers/do/d/droplet.html

packet/datasource_packet_device.go
t0mk: asnwered above

website/docs/d/device.html.markdown
t0mk: typo, thanks!

website/docs/d/device.html.markdown
t0mk: good idea, thanks!

website/docs/d/device.html.markdown
t0mk: Thanks! I blame my elementary school english teacher.

packet/datasource_packet_device.go
mmlb: ah :+1:

PR #160, opened 2019-07-25
Add datasource for organization

t0mk:
Adds datasource for Organization resource in the Packet Host.


t0mk:
@jasmingacic thanks. I will merge once I check that the acceptance tests pass.

PR #177, opened 2019-09-30
Remove incorrect statemets wrt project ssh keys

mmlb:
I believe at some point the documented behavior matched was we were doing at packet, but that was deemed that we were doing it wrong so was changed. Both the portal and kb pages document an additive behavior.


t0mk:
thanks @mmlb

PR #181, opened 2019-10-23
Set wait_for_reservation_deprovision to nil if unspecified by user

### LINKED ISSUE #180: Fix failing TestAccPacketDevice_importBasic
it fails in the TF TeamCity

```
------- Stdout: -------
=== RUN   TestAccPacketDevice_importBasic
--- FAIL: TestAccPacketDevice_importBasic (65.01s)
    testing.go:569: Step 1 error: ImportStateVerify attributes not equivalent. Difference is shown below. Top is actual, bottom is expected.
        
        (map[string]string) {
        }
        
        
        (map[string]string) (len=1) {
         (string) (len=32) "wait_for_reservation_deprovision": (string) (len=5) "false"
        }
        
FAIL
```

t0mk:
fixes #180

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -525,6 +525,11 @@ func resourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 	d.Set("network_type", device.NetworkType)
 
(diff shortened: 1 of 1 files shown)


t0mk:
tests pass (except for the flaky ones), I merge

Issue #51, opened 2018-03-05
Import-ability of existing resources


t0mk:
It would be good if existing Packet resources would be importable. Then, users could refer to a device/volume/project created out of Terraform by importing in to the TF state.

It might lower the entry barrier - user first imports existing infra, to see if TF works for her.

TF doc: https://www.terraform.io/docs/import/importability.html

I will track import-ability status here.

#45 reports lack of import in project

Currently importable

- [x] bgp_session
- [x] device
- [x] ip_attachment 
- [x] organization
- [x] port_vlan_attachment (No import test)
- [x] project
- [x] project_ssh_key
- [x] reserved_ip_block
- [ ] spot_market_request
- [x] ssh_key
- [x] vlan
- [x] volume
- [x] volume_attachment

jasmingacic:
@t0mk can you close this since it was done with #73

jasmingacic:
@t0mk I've just been looking at some imports and I noticed that most of the resources already support imports they are only missing acceptance tests. 
Do you think we could update the list to reflect actual state of the provider?

p.s. I can add missing tests

t0mk:
@jasmingacic yeah, feel free to create PR and add tests for the resources with importer. I think the tests are necessary, as some attributes break the importability, as with the `wait_for_reservation_deprovision` in `packet_device`.

nurfet-becirevic:
Enable packet_ssh_key resource import #189

nurfet-becirevic:
@t0mk , `reserved_ip_block`,  `ip_attachment`, `volume_attachment` and  `bgp_session` already support import. The acceptance tests for these resource cover import-ability as well.

t0mk:
@nurfet-becirevic Thanks, I updated the list. I also checked remaining resources and put them to the list.

t0mk:
I just found out there's a bit of a mess in 
`resource_packet_ssh_key_test.go`
and
`resource_packet_project_ssh_key_test.go`
.. project key tests are in the user key test file. I will clean it up.

displague:
Closing this issue as importers have been implemented and any furthering of this task will take place on the Equinix Metal provider.

PR #188, opened 2019-10-28
Vlan resource import

nurfet-becirevic:
Set vlan facility at import & add test


t0mk:
@nurfet-becirevic thanks, looks good. Acceptance tests run.

I've just asked for 2.6.0 release. I will merge after it's out.

jasmingacic:
This is related to #51 
@t0mk can you update the issue with what is done so we can track it.

PR #194, opened 2019-11-01
HTTP client hotfix: don't retry on HTTP 5xx

t0mk:
This is a hotfix for request retries on HTTP 5xx. 

In Packet API, HTTP 5xx responses usually indicate an issue which is not recoverable in immediate retries (even with exponential backoff).

@mmlb Manny, could you please clarify previous statement? Or could you make someone API-side-aware to explain here which HTTP errors are recoverable, if any?

### DIFF
--- a/packet/config.go
+++ b/packet/config.go
@@ -1,6 +1,11 @@
 package packet
 
 import (
+	"context"
+	"crypto/x509"
+	"net/http"
+	"net/url"
+	"regexp"
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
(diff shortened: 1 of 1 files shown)


mmlb:
That sounds correct to me (except for the unfortunatly wrong 5XX errors), let me double check with re retry-able errors. It may have to wait until Monday though.

mmlb:
> most 5XX errors are exceptions so make sense. The only one you could retry is when you get a 503 trying to create a server… as that means we are out of stock/available servers, but there is no guarantee you will get a server next time you try either.

t0mk:
@mmlb Thanks! So the conclusion is that 5xx are not recoverable with the possible exception of out-of-stock situations.

I will think a bit about the retries when out of capacity (of a plan in a facility) in the context of the TF provider and comment here again.

jasmingacic:
I believe that out of stock error should be at least 422 unprocessabile entity rather than 503 service unavailable. If we stick with 50e then the only viable solution is that we parse the error message, which isn't reliable.

t0mk:
I think there's no point retrying on 5xx HTTP errors.

I gave it a bit more thought, and I came up with 2 use-cases when it would make sense to retry:
- When the facility is out of capacity (as mentioned above). However, the retry count is in order of tens of seconds, which is negligible in comparison to just the provision time of devices in Packet. It is not negligible if we consider all the possible slots for device type in a facility, but I doubt that retrying on out-of-capacity would benefit a reasonable amount of users. If I'm provided with amount of devices in facilities and a histogram of device lifetime, I'm happy to do the math and come up with sensible retry time on out-of-capacity.
- When user hits Packet API resource throttling. This happens when user creates more than 5-6 devices at a time (for example). In this case, it would make some sense to retry, but I would argue that better solution is to either put a semaphore to device-resoruce creating function in provider code, or to control terraform paralellism - `terraform apply -parallelism=4`.

I will merge this.

jasmingacic:
Go ahead and merge it

t0mk:
acceptance tests more or less pass, I am merging

PR #195, opened 2019-11-04
Device creation rate limit

t0mk:
This PR is to limit simultaneous device creation. Packet API rate limits device resource creation, usually max 5-6 devices can be created at a time, beyond that, user will get an HTTP error.

I am aware that it's possible to control this by `-parallelism=6` in `terraform apply`, but the default of 10 is too high for Packet API at the moment. After brief discussion with Hashicorp engs in Hashicorp commiters Slack, I decided to put a semaphore to `resourcePacketDeviceCreate`, which will limit the number of simultaneous requests.

I think this will improve user experience. Not everyone is aware of `-parallelism=n`, and I think it's OK to be able to use TF without the knowledge of all the knobs and switches.

The value for the semaphore will be 6 by default, and can be set in provider config as `max_simultaneous_devices_create`, or envvar `PACKET_MAX_SIMULTANEOUS_DEVICES_CREATE`.

For review, please check the diffs for `provider.go`, `config.go` and `resource_packet_device.go`. Also, feel free to suggest different naming.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/mattn/go-colorable v0.1.1 // indirect
 	github.com/packethost/packngo v0.2.1-0.20191030123207-e34c34ef3c41
 	github.com/vmihailenco/msgpack v4.0.1+incompatible // indirect
+	golang.org/x/sync v0.0.0-20190423024810-112230192c58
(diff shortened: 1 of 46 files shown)


t0mk:
I put this on hold for now, no revied needed.

PR #197, opened 2019-11-05
fix acceptance tests for Layer 2 networking (ports and VLANS)

### LINKED ISSUE #182: vlan resource tests fail
Failing acceptance tests

Following tests notoriously fails in Hashicorp Teamcity:

They are all the L2 ports assigning tests.

TestAccPacketPortVlanAttachment_Hybrid 
TestAccPacketPortVlanAttachment_HybridMultipleVlans
TestAccPacketPortVlanAttachment_L2Bonded
TestAccPacketPortVlanAttachment_L2Individual
TestAccPacketPortVlanAttachment_L2Native

The error is always

```
the facility dfw2 has no provisionable s1.large.x86 servers matching your criteria
        
          on /opt/teamcity-agent/temp/buildTmp/tf-test955797134/main.tf line 6:
          (source code not available)
     
```

The facility must be changed or found out during runtime based on capacity.

t0mk:
The Acceptance tests for VLAN attachment notoriously fail (#182). 

They fail either because facility `dfw2` is out of capacity of L2-capacble device plans, or there are timeouts.

This PR will switch the facility to "nrt1", hopefully making the tests succeed for some time. This is obviously just a temporary solution, but it's good to see the VLAN tests succeed again, after quite some changes in the provider.

Permanent solution would be to get the capacity via a datasource and run the test devices where the capacity is not "unavailable".

closes #182 temporarily

### DIFF
--- a/packet/resource_packet_port_vlan_attachment_test.go
+++ b/packet/resource_packet_port_vlan_attachment_test.go
@@ -21,7 +21,7 @@ resource "packet_project" "test" {
 resource "packet_device" "test" {
   hostname         = "test"
   plan             = "s1.large.x86"
-  facilities       = ["dfw2"]
+  facilities       = ["nrt1"]
   operating_system = "ubuntu_16_04"
   billing_cycle    = "hourly"
   project_id       = "${packet_project.test.id}"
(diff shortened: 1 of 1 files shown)


t0mk:
The port_vlan_attachment tests pass now, I am merging.

Issue #206, opened 2019-12-03
bgp resource fails to provision IPV6


coryschwartz:
### Affected Resource(s)
Please list the resources as a list, for example:
- packet_bgp_session

If this issue appears to affect multiple resources, it may be an issue with Terraform's core, so please mention this.

### Terraform Configuration Files
```hcl
resource "packet_bgp_session" "gateway_bgp_ipv4" {
  device_id = "${packet_device.gateway.id}"
  address_family = "ipv4"
}

resource "packet_bgp_session" "gateway_bgp_ipv6" {
  device_id = "${packet_device.gateway.id}"
  address_family = "ipv6"
}
```

### Debug Output

The first run
```
Error: Error applying plan:
2 errors occurred:
	* module.gateway_stage1.packet_bgp_session.gateway_bgp_ipv4: 1 error occurred:
	* packet_bgp_session.gateway_bgp_ipv4: Oh snap, something went wrong! We've logged the error and will take a look - please reach out to us if you continue having trouble.
	* module.gateway_stage2.packet_bgp_session.gateway_bgp_ipv4: 1 error occurred:
	* packet_bgp_session.gateway_bgp_ipv4: Oh snap, something went wrong! We've logged the error and will take a look - please reach out to us if you continue having trouble.
```

and then subsequent runs:
```
Error: Error applying plan:
2 errors occurred:
	* module.gateway_stage2.packet_bgp_session.gateway_bgp_ipv4: 1 error occurred:
	* packet_bgp_session.gateway_bgp_ipv4: Address family already has session with ipv4
	* module.gateway_stage1.packet_bgp_session.gateway_bgp_ipv4: 1 error occurred:
	* packet_bgp_session.gateway_bgp_ipv4: Address family already has session with ipv4
```

### Expected Behavior
I should be able to have BGP sessions for IPV6 and IPV4

### Actual Behavior
IPV4 sessions are created, IPV6 sessions need to be enabled using the packet web UI.

### Steps to Reproduce
Please list the steps required to reproduce the issue, for example:
1. `terraform apply`

### References
I think I might know what's going on here. While the address_family is read from the configuration, it is ignored and only "ipv4" is used when creating the session.

https://github.com/terraform-providers/terraform-provider-packet/blob/master/packet/resource_packet_bgp_session.go#L55

t0mk:
@coryschwartz thanks for the report. I'm sad that this went through, it's my fault, sorry.

[...]

PR #207, opened 2019-12-03
Use the address family provided, rather than just ipv4

coryschwartz:
This patch would create the BGP session using the provided address
family.

The tests provided in this patch would not actually catch the error,
since they since they jsut test for valid configuraiton, not the objects
used when the session is created.

By the way, I have not tested this patch in any way.

Maybe fixes https://github.com/terraform-providers/terraform-provider-packet/issues/206


t0mk:
@coryschwartz I will try to improve the tests tomorrow.

PR #215, opened 2019-12-18
Improve L2 acceptance test

t0mk:
VLAN-port attachment tests are failing in Hashicorp CI, mostly because of full capacity.

This PR will take advantace of the dynamic "facilitties" field in packet_device so that the L2 test device has better chance to be deployed.

### DIFF
--- a/packet/resource_packet_port_vlan_attachment_test.go
+++ b/packet/resource_packet_port_vlan_attachment_test.go
@@ -24,20 +24,20 @@ resource "packet_device" "test" {
   facilities       = ["nrt1"]
   operating_system = "ubuntu_16_04"
   billing_cycle    = "hourly"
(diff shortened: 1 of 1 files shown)


displague:
This seems like an easy merge if rebased.

Would it really help though? Are the tests simple enough that we only require a facility with enough of one type of device?

If we need more capacity checking that `facility = "any"` can provide, we might need a new data source that can do more advanced capacity checking (based on required facility features and having at least count devices for multiple device types) and return the first matching facility.

t0mk:
@displague I think so far, device types in tests don't mix. I would keep it that way in order to keep allocation simpler. At the moment, I don't see any advantage in mixing device types in acceptance tests.

But maybe we can get back to this only if we will hit the capacity problem again.

PR #233, opened 2020-05-01
Update spot_market_request.html.markdown

Miouge1:
Correct Packet name: Packet not Packing :)


t0mk:
thanks for the fix @Miouge1 :)

PR #235, opened 2020-05-11
remove public_ipv4_subnet_size

t0mk:
…  in favor of `ip_address` attribute and referencing to a reserved IP block


t0mk:
most of tests pass, I merge

PR #239, opened 2020-07-09
fix conversion from nil during WaitForState loops

displague:
As [reported in Slack](https://packetcommunity.slack.com/archives/CU7HC3RJP/p1594207906242100), `nil` responses were being converted to `string`, causing the following
error pattern:

```
[DEBUG] GET https://api.packet.net/devices/{uuid}?include=project,facility
[TRACE] Waiting 10s before next try
[WARN] WaitForState timeout after 1h0m0s
[WARN] WaitForState starting 30s refresh grace period

panic: interface conversion:
interface {} is nil, not string
```

---

*The terraform-plugin-sdk code triggering this seems related ([based on EC2 comments](https://github.com/hashicorp/terraform-plugin-sdk/blob/master/helper/resource/state.go#L16)) to the AWS provider, which has [evolved the pattern](https://github.com/terraform-providers/terraform-provider-aws/blob/711107a93685d175c3a00dbee50ac5ee26d8c085/aws/internal/service/workspaces/waiter/waiter.go#L83-L89).*

### DIFF
--- a/packet/helpers_device.go
+++ b/packet/helpers_device.go
@@ -187,9 +187,14 @@ func waitForDeviceAttribute(d *schema.ResourceData, targets []string, pending []
 		Delay:      10 * time.Second,
 		MinTimeout: 3 * time.Second,
 	}
-	attrval, err := stateConf.WaitForState()
 
-	return attrval.(string), err
+	attrValRaw, err := stateConf.WaitForState()
+
+	if v, ok := attrValRaw.(string); ok {
+		return v, err
+	}
+
+	return "", err
 }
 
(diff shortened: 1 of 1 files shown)


t0mk:
Hi @displague, thanks for the PR. It looks that this PR will do no harm. I will merge it. Do you need this to be released early?

Do you know any more detail of when this was happening? I guess it was a  wait for some vmware device?

displague:
Hey @t0mk, yes, this problem was reported in Slack (link in the description).

I've also created https://github.com/packet-labs/google-anthos/issues/80 to track this problem, should this PR not resolve it.

PR #246, opened 2020-08-20
Updates documentation to reflect new repo URL

jmarhee:
Revises github.com/terraform-provider references to github.com/packethost in:
* README.md
* go.mod/go.sum
* import paths
* documentation


displague:
Actually, @jmarhee, can we also update the go module references too (go.mod, imports, etc).  
`go mod tidy` may be needed after that.

go.mod
```diff
@@ -16,4 +14,6 @@ require (
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
 
+replace github.com/packethost/terraform-provider-packet => ../terraform-provider-packet
```
displague, line 17: this seems out of place

Issue #249, opened 2020-08-28
add a data resource for getting ssh keys


displague:
There is currently no way to query and access an existing Packet Project SSH Keys through Terraform.

Create a data resource for project ssh keys.

displague:
https://github.com/packethost/packngo/pull/190 would enable a three-field `search` parameter for this data resource.
The existing `SSHKeyServiceOp.ProjectList` in packngo is sufficient for this data source.

t0mk:
@displague I'll take a look during the weekend or on Monday

gianarb:
Not sure about why this one is assigned to me :)

displague:
This resource was included in the Packet provider and is present in the Equinix Metal provider.

PR #250, opened 2020-08-28
add datasource packet_project_ssh_key

displague:
Implements #249 

Signed-off-by: Marques Johansson <marques@packet.com>

### DIFF
--- a/packet/datasource_packet_project_ssh_key.go
+++ b/packet/datasource_packet_project_ssh_key.go
@@ -1,0 +1,125 @@
+package packet
+
+import (
+	"fmt"
+	"path"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"github.com/packethost/packngo"
+)
+
+func dataSourcePacketProjectSSHKey() *schema.Resource {
+	return &schema.Resource{
+		Read: dataSourcePacketProjectSSHKeyRead,
+		Schema: map[string]*schema.Schema{
+			"search": {
+				Type:         schema.TypeString,
+				Description:  "The name, fingerprint, id, or public_key of the SSH Key to search for in the Packet project",
+				Optional:     true,
+				ValidateFunc: validation.NoZeroValues,
+			},
+			"id": {
+				Type:         schema.TypeString,
+				Description:  "The id of the SSH Key",
+				Optional:     true,
+				ValidateFunc: validation.NoZeroValues,
+				Computed:     true,
+			},
+			"project_id": {
+				Type:         schema.TypeString,
+				Description:  "The Packet project id of the Packet SSH Key",
+				Required:     true,
+				ValidateFunc: validation.NoZeroValues,
+			},
+			"name": {
+				Type:        schema.TypeString,
+				Description: "The label of the Packet SSH Key",
+				Computed:    true,
+			},
+			"public_key": {
+				Type:        schema.TypeString,
+				Description: "The public SSH key that will be authorized for SSH access on Packet devices provisioned with this key",
+				Computed:    true,
+			},
+			"fingerprint": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"created": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"updated": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"owner_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+		},
+	}
+}
+
+func dataSourcePacketProjectSSHKeyRead(d *schema.ResourceData, meta interface{}) error {
+	client := meta.(*packngo.Client)
+
+	search := d.Get("search").(string)
+	id := d.Get("id").(string)
+	projectID := d.Get("project_id").(string)
+
+	if id == "" && search == "" {
+		return fmt.Errorf("You must supply either search or id")
+	}
+
+	var (
+		key        packngo.SSHKey
+		searchOpts *packngo.SearchOptions
+	)
+
+	if search != "" {
+		searchOpts = &packngo.SearchOptions{Search: search}
+	}
+	keys, _, err := client.Projects.ListSSHKeys(projectID, searchOpts)
+
+	if err != nil {
+		err = fmt.Errorf("Error listing project ssh keys: %s", friendlyError(err))
+		return err
+	}
+
+	for i := range keys {
+		// use the first match for searches
+		if search != "" {
+			key = keys[i]
+			break
+		}
+
+		// otherwise find the matching ID
+		if keys[i].ID == id {
+			key = keys[i]
+			break
+		}
+	}
+
+	if key.ID == "" {
+		// Not Found
+		return fmt.Errorf("Project %q SSH Key matching %q was not found", projectID, search)
+	}
+
+	ownerID := path.Base(key.Owner.Href)
+
+	d.SetId(key.ID)
+	d.Set("name", key.Label)
+	d.Set("public_key", key.Key)
+	d.Set("fingerprint", key.FingerPrint)
+	d.Set("owner_id", ownerID)
+	d.Set("created", key.Created)
+	d.Set("updated", key.Updated)
+
+	if key.Owner.Href[:10] == "/projects/" {
+		d.Set("project_id", ownerID)
+	}
+
+	return nil
+}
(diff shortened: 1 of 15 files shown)


t0mk:
@displague I will take a look during the weekend or on Monday

displague:
The test isn't passing yet.

```
PACKNGO_DEBUG=1 TESTARGS="-run TestAccPacketProjectSSHKeyDataSource -count=1"  make testacc
```

packet/datasource_packet_project_ssh_key.go
```diff
@@ -0,0 +1,102 @@
...
+
+	keys, _, err := client.SSHKeys.ProjectList(projectID)
+
+	if err != nil {
+		err = fmt.Errorf("Error listing project ssh keys: %s", friendlyError(err))
+		return err
+	}
+
+	for _, testKey := range keys {
+		if testKey.Label == search || testKey.Key == search || testKey.ID == search || testKey.FingerPrint == search {
```
t0mk, line 77: I think using general field (`search`) doesn't fit in the tf datasource concept. There should be concrete fields usable for search for those particular fields. I would limit the search for only
* `label` (no need to call it a `name`, it's `label` in Packet API)
* `fingerprint`

^ is the norm IIUC, see https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/data-sources/account_ssh_key

I would drop
* search by `public_key` (can't think of a scenario where it would be useful). 
* search by `id`, which can be better used for `terraform import` I think. But I don't insist on this. If you keep search by ID, the field should be `project_ssh_key_id`.

t0mk (changes requested):
Looks good overall, I just disagree with the `search` field. 

Also, think about uniting user and project keys in the datasource, it can save some code. I guess you've seen the `packet_ssh_key` resource.

packet/datasource_packet_project_ssh_key.go
displague: > label (no need to call it a name, it's label in Packet API)

I felt that way too, but in the Terraform provider the field is already being used as `name`, and not `label`.  I was trying to stay consistent to the Terraform provider.

> There should be concrete fields usable for search for those particular fields
I was taking advantage of an actual `search` query parameter provided by the API: _"Search by key, label, or fingerprint"_

I have a packngo PR that implements this with the API parameter: https://github.com/packethost/packngo/pull/190

To be faithful to the API and the eventual drop-in use of packethost/packngo#190, I should move `ID` to a separate datasource parameter.

>  If you keep search by ID, the field should be project_ssh_key_id.

I wanted to double-check this, I think the convention here is to match the field name. The AWS provider uses `id`, for example: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/data-sources/eip (on many of the data sources). The GCP provider uses `something_id`.

> I think using general field (search) doesn't fit in the tf datasource concept

I think that depends on the capabilities of the provider, for example, this GCP datasource takes "filter" as a parameter:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/data-sources/projects#example-usage---searching-for-projects-about-to-be-deleted-in-an-org














> search by public_key (can't think of a scenario where it would be useful).

👍  If I keep search, because it exists in the API, we won't need a separate `public_key`.  If we do need one, someone can request it.

t0mk: Makes sense with the ID attr naming, I'm always up for less text, but with these little things in TF, convention seems to be in the eye of the beholder:
https://registry.terraform.io/providers/scaleway/scaleway/latest/docs/data-sources/instance_image

.. I look at scaleway because they put a lot of effort in their TF provider, and I see it better align-fit to Packet than aws or gc. In effect, I don't care either way, but I would keep it consistent with the other datasoruces in the provider at a release time. I.e. let's change it either way, but for all DSs at a time. And maybe in a separate PR.

I wasn't aware about the `search` param in the API until yesterday. Why not wait till the packethost/packngo#190 PR is in and then drop fingerptint, public key and label, and only keep the ID.

I reallt don't see the point of `filter` once packethost/packngo#190 is in. You mean there would be 2 llayers of filtering, once in the API and then in TF provider?

t0mk:
I will check the test.

packet/datasource_packet_project_ssh_key_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/acctest"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+)
+
+func TestAccPacketProjectSSHKeyDataSource_Basic(t *testing.T) {
+	datasourceName := "data.packet_project_ssh_key.foobar"
+
+	publicKeyMaterial, _, err := acctest.RandSSHKeyPair("")
+	keyName := acctest.RandomWithPrefix("tfacc-project-key")
```
t0mk, line 16: maybe move this one line up, so that it's clear that the error check is for RansdSSHKeyPair

packet/datasource_packet_project_ssh_key_test.go
```diff
@@ -0,0 +1,127 @@
...
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						datasourceName, "name", keyName),
+					resource.TestCheckResourceAttr(
+						datasourceName, "public_key", publicKeyMaterial),
+				),
+			},
+			{
+				Config:      testAccPacketProjectSSHKeyDataSourceConfig_noKey(keyName, publicKeyMaterial),
+				ExpectError: regexp.MustCompile(keyName + " was not found"),
```
t0mk, line 70: The test is failiing with 2 errors. First is

```
        errors during follow-up refresh: Project "788ede5c-35e0-4d2f-95db-21897a4e22ae" SSH Key matching "tfacc-project-key-1049025380679756230" was not found 0
        
        To match:
        
        tfacc-project-key-1049025380679756230 was not found
```

Because the regexp doesn't match the error. There's a double-quote char after the key ID.

The regexp should be `regexp.MustCompile(keyName + "\" was not found")`

t0mk: [WIP]
I was not able to debug the other error, namely
```
inset error
```

The only fix I found to work was to separate the device removal 
```go
func testAccPacketProjectSSHKeyDataSourceConfig_noKey2(keyName string) string {
    config := fmt.Sprintf(`
data "packet_project_ssh_key" "foobar" {
    search = "%s"
    project_id = packet_project.test.id
}

resource "packet_project" "test" {
    name = "%s"
}`, keyName, keyName)
    return config
}

func TestAccPacketProjectSSHKeyDataSource_NoKey(t *testing.T) {
    keyName := acctest.RandomWithPrefix("tfacc-project-key")

    rekn := regexp.MustCompile(keyName + "\" was not found")

    resource.Test(t, resource.TestCase{
        PreCheck:     func() { testAccPreCheck(t) },
        Providers:    testAccProviders,
        //CheckDestroy: testAccCheckPacketSSHKeyDestroy,
        Steps: []resource.TestStep{
            {
                Config:      testAccPacketProjectSSHKeyDataSourceConfig_noKey1(keyName),
            },
            {
                Config:      testAccPacketProjectSSHKeyDataSourceConfig_noKey2(keyName),
                ExpectError:rekn,
            },
        },
    })
}

```

packet/datasource_packet_project_ssh_key.go
displague: I was only using Google's `filter` field as an example.  The Packet param would be called `search` to match the query parameter name that the Packet API uses. Yes, I do mean that there would be two layers, that the Terraform client would be implementing "search". If packethost/packngo#190 is merged first I will gladly remove the logic here.

packet/datasource_packet_project_ssh_key_test.go
displague: https://github.com/terraform-providers/terraform-provider-aws/pull/7370 was the trick we needed here (an empty config following an errored config so that the destroy step will succeed).

I also added `PreventPostDestroyRefresh: true,` which may not be needed with the empty config trick.

The tests are passing now.

t0mk (approved):
The tests run. Good job fixing them.

packet/datasource_packet_project_ssh_key_test.go
t0mk: PACKNGO_DEBUG=1 never worked for me in the provider testing, I've always been using TF_LOG=DEBUG to see the HTTP reqs/responses.

But I tried now with PACKNGO_DEBUG=1 and I didn't notice a delay.

I had troubles running the tests in the provider and in packngo when using go 1.15. I always had to switch to 1.13 (from go.mod of both of the projects), to test the code.

We should probably keep the Go version same in both of the projects.

PR #261, opened 2020-09-22
add GH workflow for acceptance testing

displague:
This adds acceptance testing back into the PR workflow using Github Actions.
* [x] Requires a `PACKET_AUTH_TOKEN` to be added to https://github.com/packethost/terraform-provider-packet/settings/secrets

Previously, this was performed on Hashicorp managed servers in TeamCity.

I would like to restrict the PR builds to ones that have received a label of preliminary acceptance (something like "ok-to-test"), but I'd be ok with full acceptance testing being introduced before then (if we disable build-on-PR and only use build-on-push to the default branch).

I'm not sure if we want to add these environment variables to the build at this time, we can always add them later:

```
TF_SCHEMA_PANIC_ON_ERROR: "1" # I think this will be the default in TF sdk2
TF_LOG: "DEBUG"
```

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -1,0 +1,76 @@
+name: Acceptance Tests
+on:
+  pull_request:
+    paths-ignore:
+      - 'LICENSE'
+      - '**.md'
+      - 'website/**'
+      - 'docs/**'
+  push:
+    paths-ignore:
+      - 'LICENSE'
+      - '**.md'
+      - 'website/**'
+      - 'docs/**'
+  schedule:
+    - cron: '0 13 * * *'
+jobs:
+
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.2
+      with:
+        go-version: '1.15'
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2.3.2
+
+    - name: Get dependencies
+      run: |
+        go mod download
+    - name: Build
+      run: |
+        go build -v .
+  test:
+    name: Matrix Test
+    needs: build
+    runs-on: ubuntu-latest
+    timeout-minutes: 240
+    strategy:
+      fail-fast: false
+      matrix:
+        version:
+          - stable
+        terraform:
+          - '0.12.29'
+          - '0.13.3'
+    steps:
+
+    - name: Set up Go
+      uses: actions/setup-go@v2.1.2
+      with:
+        go-version: '1.15'
+      id: go
+
+    - name: Check out code into the Go module directory
+      uses: actions/checkout@v2.3.2
+
+    - name: Get dependencies
+      run: |
+        go mod download
+
+    - name: TF acceptance tests
+      timeout-minutes: 120
+      env:
+        TF_ACC: "1"
+        TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
+
+        PACKET_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
+      run: |
+        go test -v -cover ./packet
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -25,7 +25,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2.1.2
       with:
-        go-version: '1.15'
+        go-version: '1.13'
       id: go
 
     - name: Check out code into the Go module directory
@@ -55,7 +55,7 @@ jobs:
     - name: Set up Go
       uses: actions/setup-go@v2.1.2
       with:
-        go-version: '1.15'
+        go-version: '1.13'
       id: go
 
     - name: Check out code into the Go module directory
@@ -73,4 +73,4 @@ jobs:
 
         PACKET_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
       run: |
-        go test -v -cover ./packet
+        go test -v -cover -run=TestAccPacketProject_Basic ./packet
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -76,7 +76,7 @@ jobs:
 
         PACKET_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
       run: |
-        go test -v -cover ./packet
+        go test -v -cover -parallel 4 -timeout 120m ./packet
     - name: Sweeper
       if: ${{ always() }}
       env:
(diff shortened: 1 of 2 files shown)


displague:
I attempted to use this build with a secret here: https://github.com/displague/terraform-provider-packet/runs/1151431036?check_suite_focus=true . 

It timed out after 10m. I don't see anything in Terraform or the GH Actions that should trigger this 🤔 .
@t0mk Are there extra environment variables configured in TeamCity (I don't have access to this)?

t0mk:
hey @displague , I will take a look tomorrow.

displague:
I've added the secrets. I was surprised that this PR already triggered the AccTest builds, perhaps since I'm a contributor in the same org.

t0mk:
@displague I would start with fixing Go version in the acc test agent to 1.13, like it's in go.mod. Or perhaps parsing go.mod for the Go verison (in case we'd like to update).

t0mk:
@displague I set Go version in the Actions Workflow to Go 1.13 (like it's in go.mod), and I limited the tests to only `-run=TestAccPacketProject_Basic`, so that not all the tests are run, and we can debug/iterate faster.

t0mk:
hmm, even the single test is still failing with
```
=== RUN   TestAccPacketProject_Basic
--- FAIL: TestAccPacketProject_Basic (0.00s)
    provider_test.go:33: PACKET_AUTH_TOKEN must be set for acceptance tests
FAIL
```

I can't see the secrets in the settings (I'm only Contributor), so I can't check the PACKET_AUTH_TOKEN there.

t0mk:
@displague It seems that the secrets are expanded only in PRs from local branches, see successful acc test runs in a PR that I've created from a branch pushed to this repo:

https://github.com/packethost/terraform-provider-packet/pull/262/checks

displague:
@t0mk This puts the honus on the reviewer to run acctests before merging contributor PRs 🤔

We could accept that and then use release branches (on packethost) to trigger E2E tests.  We could also push our own PRs (the majority of the PRs) directly to `packethost` so they do get full testing. 

This seems ok to me. What are your thoughts?

displague:
The other outstanding problem is the unexpected 10m timeout.

jmarhee:
> @t0mk This puts the honus on the reviewer to run acctests before merging contributor PRs 🤔
> 
> We could accept that and then use release branches (on packethost) to trigger E2E tests. We could also push our own PRs (the majority of the PRs) directly to `packethost` so they do get full testing.
> 
> This seems ok to me. What are your thoughts?

This seems a reasonable reviewer expection (and probably necessary additional quality control)  to me as well.

t0mk:
@displague 

> @t0mk This puts the honus on the reviewer to run acctests before merging contributor PRs thinking

I think that running and checking ALL the acceptance tests __locally__ is too much burden for a reviewer, it can take 40 minutes to run them all, plus the reviewer would have to manually report which went wrong, and guess if a test fail was flake or not, etc.

I would suggest, on the rare even of outside contribution, that reviewer will check the code, run only relevant acceptance tests, and then create a copy of the PR in this repo (maybe called `acctest_<pr_name>`), which will trigger __all__ the acceptance tests, and then we will see all the test results on github.

> We could accept that and then use release branches (on packethost) to trigger E2E tests. 

Not sure if testing in release branches will work e.g. when you have 2 big PRs opened by outsiders and you will want to see the whole acc test suite results for both of them. But there's not too many major contributions from outsiders, so perhaps this is not a scenario for which we'd need to plan for. 

> We could also push our own PRs (the majority of the PRs) directly to packethost so they do get full testing.

I agree with this. Let's create own PR branches in this repo. I would leave testing outside contibutions on case-by-case basis.

I think ^ is the best we can do with CI built from Github Actions. Another option would be for Packet to use some external CI system, just like when Hashicorp was using the Teamcity. Then, it would be possible to trigger CI runs on branches manually.

.github/workflows/acctest.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    - name: TF acceptance tests
+      timeout-minutes: 120
+      env:
+        TF_ACC: "1"
+        TF_ACC_TERRAFORM_VERSION: ${{ matrix.terraform }}
+
+        PACKET_AUTH_TOKEN: ${{ secrets.PACKET_AUTH_TOKEN }}
+      run: |
+        go test -v -cover -run=TestAccPacketProject_Basic ./packet
```
t0mk, line 76: before merge, remember to remove `-run=TestAccPacketProject_Basic`, so that it runs all the tests.

displague:
@t0mk The 10m timeouts are a problem. I'm not seeing the sweepers running either. All that said, let's merge this and figure it out with this base in place.

t0mk:
@displague the sweepers were run separatey in HC Teamcity. Before and after the acc tests there was a shellscript task

```
go test ./%PROVIDER_PATH%/%PROVIDER_NAME% -v -sweep="%SWEEPER_REGIONS%"
```

t0mk:
@displague Can you please point me to the 10-minute timeout? I am not sure what you mean.

displague:
@tomk pointed out elsewhere:
> Could the 10 minutes be just default go test timeout?

Yes, that's probably it. Because this PR is not running `make testacc` and is reproducing the `go test` command, it has differed in that way.

displague:
The test failed, but not because of a timeout. It failed because I added `TF_SCHEMA_PANIC_ON_ERROR=1` and this provider is not ready for that.

```
panic: Invalid address to set: []string{"type"}
goroutine 4107 [running]:
[...]
	/home/runner/work/terraform-provider-packet/terraform-provider-packet/packet/resource_packet_reserved_ip_block.go:185 +0x4ee
```

We can take that on in a separate PR. We'll need to do this eventually for tf-sdk-2.

displague:
I'm prepared to merge this if https://github.com/displague/terraform-provider-packet/actions/runs/282947594 passes. 🤞

displague:
I'm prepared to merge this if https://github.com/displague/terraform-provider-packet/actions/runs/282947594 passes. 🤞

displague:
The following test failed:
```
--- FAIL: TestAccPacketSpotMarketRequest_Basic (409.26s)
    testing.go:569: Step 0 error: After applying this step and refreshing, the plan was not empty:
```

I think we can address this refresh related (or diff jitter) test separately.

The other result that was unexpected was that the sweeper failed to delete a device that was still provisioning.

I'd like for the sweeper to deal with these things immediately, but this is exactly the type of problem that we have a scheduled test and sweep job phase for. It will eventually resolve.  It should leave at most one test build of resources behind.

```
2020/10/01 20:19:14 [ERR] error running (packet_organization): Error deleting device DELETE https://api.packet.net/devices/25fb6701-a45e-40c9-8ace-bc0af69b6855: 422 Cannot delete a device while it is provisioning 
FAIL	github.com/packethost/terraform-provider-packet/packet	12.970s
FAIL
Error: Process completed with exit code 1.
```

t0mk (approved):
@displague nice, most of the tests passed!

The spot market req test was always flaky, so it's alright. The sweep err is sad, but OK. We could also try to waitForDeviceAttribute("state", "active") there.

It would be nice to see the test result in a table, so that it would be obvious on the first look which tests passed and (more importantly) which tests failed. Is this somehow possible?

But feel free to merge.

PR #271, opened 2020-10-07
tests: don't attempt acctest on pull requests

displague:
Pull requests which are based on `origin/` branches will be `acctest`'ed
because of the [`push` rule](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#push). The [`pull_request` rule](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request) can not work because
GH actions does not pass secrets to pull requests from forks.

Signed-off-by: Marques Johansson <marques@packet.com>

### DIFF
--- a/.github/workflows/acctest.yml
+++ b/.github/workflows/acctest.yml
@@ -1,11 +1,5 @@
 name: Acceptance Tests
 on:
-  pull_request:
-    paths-ignore:
-      - 'LICENSE'
-      - '**.md'
-      - 'website/**'
-      - 'docs/**'
   push:
     paths-ignore:
       - 'LICENSE'


displague:
I wonder if https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target would permit secrets?  I'm ok with not testing pull_requests from forks, for now. (I just have to get in the habit of pushing branches to origin 🤦 )

t0mk (commented):
@displague it makes sense NOT to trigger CI on pull_request event. The secrets from packethost/t-f-p are not available to PRs from forked repos, so, in general, running CI has no point there.

The pull_request_target trigger makes more sense because it will theoretically work for PRs from local branches, and for PRs from forks with configured PACKET_AUTH_TOKEN (like yours). It will fail for ousider PRs, but that's OK maybe.

As for trigger on push event, it seems a bit too expensive to run CI suite on every commit. I would agree if we'd make a small subset of "fast" tests, and those would be run on each commit. Maybe those that just touch the API and don't need to wait, like creating project, volume, user, just these small things.

I like running the CI only on manual trigger:
https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow
.. if possible we'd set it that only contributors can trigger it, but I'm not sure if this can be done. Also, it would only make sense to trigger tests for local branches, so that they can access the secrets.

displague:
> As for trigger on push event, it seems a bit too expensive to run CI suite on every commit.

I thought this was only running for every HEAD commit on `master` (good) and origin branches (good, our alternative to pull request builds, which will build every origin pull request (branch HEAD)). 

It would be nice to only test local branches that have a label associated with the PR. I agree that https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#workflow_dispatch sounds interesting, but it is not clear to me how we would accomplish this (I'm willing to hold out for more examples).

Branch naming could also filter this down: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-ignoring-branches-and-tags.

These are all things we could do in the future. Do you see any reason why this PR should be blocked?

t0mk:
@displague feel free to merge

PR #273, opened 2020-10-14
update packngo to 0.4.1 (new API endpoint api.equinix.com/metal/v1)

displague:
Updates the Packngo version to one that is aware of the new Equinix Metal API endpoint at (api.equinix.com/metal/v1).

Users will want to revise any network filters to permit the new endpoint.

Nothing else changes about packngo in this version change (including PACKNGO_DEBUG and the default auth token environment variables).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## 3.1.0 (Unreleased)
 
+BREAKING CHANGES:
+- packngo updated to v0.4.1, changing the API endpoint from api.packet.net to api.equinix.com/metal/v1
+
 FEATURES:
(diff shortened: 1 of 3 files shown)


detiber (approved):

t0mk:
@displague this would be a perfect PR to test out the acceptance test suite over here at Github.

displague:
@t0mk I'll merge this in a create a prerelease tag for it, this will be a good time to test that feature of the registry.

Terraform should accept `v3.1.0-alpha1` as the tag name, I'd like to do the same with the packngo v0.5.0 release (alpha2).

Issue #264, opened 2020-10-02
Tests fail with `TF_SCHEMA_PANIC_ON_ERROR=1`


displague:
The tests fail with `TF_SCHEMA_PANIC_ON_ERROR=1`. This must be overcome for TF SDK2 compatibility.

Terraform returns errors when a `Set` is called with an invalid value for the field type.  These errors are generally not handled by providers. With this TF_SCHEMA_PANIC_ON_ERROR flag set, those errors are returned as a panics.  This is useful in testing to ensure that all `Set` calls are effective.

```
panic: Invalid address to set: []string{"type"}
goroutine 4107 [running]:
[...]
	/home/runner/work/terraform-provider-packet/terraform-provider-packet/packet/resource_packet_reserved_ip_block.go:185 +0x4ee
```

_Originally posted by @displague in https://github.com/packethost/terraform-provider-packet/pull/261#issuecomment-702333804_

PR #274, opened 2020-10-14
adds setMap to check errors when setting IP block data

displague:
This is for consideration of #264.

This may make `d.Set` blocks easier to read, if implemented everywhere, and it would ensure that `d.Set` errors are being checked and reported.  Without a function like `setMap` we could end up with hundreds of `if err := d.Set(...); err != nil { return err }` blocks added to address #264.

One potential down-side of this, is that Terraform plugins for VS Code (and perhaps linters?) that are smart about `d.Set`, may not be able to provide guidance using this approach.  I don't know if that is an actual capability of linters and IDE-plugins.

`github.com/hashicorp/go-multierror` is used to collect and collate the `d.Set` errors.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,13 +4,13 @@ require (
 	github.com/apparentlymart/go-dump v0.0.0-20190214190832-042adf3cf4a0 // indirect
 	github.com/aws/aws-sdk-go v1.22.0 // indirect
 	github.com/hashicorp/errwrap v1.0.0
+	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/go-retryablehttp v0.6.6
 	github.com/hashicorp/terraform-plugin-sdk v1.0.0
 	github.com/mattn/go-colorable v0.1.1 // indirect
 	github.com/packethost/packngo v0.3.0
 	github.com/vmihailenco/msgpack v4.0.1+incompatible // indirect
-	golang.org/x/net v0.0.0-20191004110552-13f9640d40b9 // indirect
-	golang.org/x/sys v0.0.0-20191022100944-742c48ecaeb7 // indirect
+	golang.org/x/net v0.0.0-20200822124328-c89045814202 // indirect
 	gopkg.in/yaml.v2 v2.2.8 // indirect
 )
 
--- a/packet/errors.go
+++ b/packet/errors.go
@@ -2,7 +2,11 @@ package packet
 
 import (
 	"net/http"
+	"sort"
 	"strings"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 
 	"github.com/packethost/packngo"
 )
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -4,13 +4,14 @@ require (
 	github.com/apparentlymart/go-dump v0.0.0-20190214190832-042adf3cf4a0 // indirect
 	github.com/aws/aws-sdk-go v1.22.0 // indirect
 	github.com/hashicorp/errwrap v1.0.0
+	github.com/hashicorp/go-multierror v1.0.0
 	github.com/hashicorp/go-retryablehttp v0.6.6
 	github.com/hashicorp/terraform-plugin-sdk v1.0.0
+	github.com/kisielk/errcheck v1.4.0 // indirect
```
displague, line 10: This was added by `make errcheck` - I should remove it.

packet/errors.go
```diff
@@ -61,3 +65,28 @@ type ErrorResponse struct {
...
+
+// setMap sets the map of values to ResourceData, checking and returning the
+// errors. Typically d.Set is not error checked. This helper makes checking
+// those errors less tedious. Because this works with a map, the order of the
+// errors would not be predictable, to avoid this the errors will be sorted.
+func setMap(d *schema.ResourceData, m map[string]interface{}) error {
+	errs := &multierror.Error{}
+	for key, v := range m {
+		var err error
+		if f, ok := v.(setFn); ok {
```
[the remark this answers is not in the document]
displague: @detiber created a simple repro for this: https://play.golang.org/p/vqglaCZLtp_R 

Playing with that a little, I found that I could 
use the raw type (what is that actually called?) https://play.golang.org/p/hfDdSpfgzk_K
or I could use `=` in the type
https://play.golang.org/p/Wo7jdZKIJ3m

I've updated the type to use `=`.
Thanks, @detiber!

go.mod
displague: ran `go mod tidy`

displague:
Tests are running at https://github.com/displague/terraform-provider-packet/actions/runs/306837661 (I should have made this PR against origin).

displague:
The relevant tests passed.

detiber (approved):

t0mk:
@displague 

I think a normal err check for the d.Sets would be enough, but this makes the code shorter and moved err checks and attr assignments to once place. 

It looks good, the tests for reserved IP block succeed. Would be good to have a test to see that the error handling indeed works, but as the checks are done for data returned from the API, it's probably not trivial.

packet/resource_packet_reserved_ip_block.go
```diff
@@ -170,40 +170,49 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
...
 			return fmt.Errorf("Strange (too small) CIDR prefix: %d", reservedBlock.CIDR)
 		}
-		d.Set("quantity", 1<<uint(bits))
+		quantity = 1 << uint(bits)
 	}
-	d.Set("project_id", path.Base(reservedBlock.Project.Href))
-	d.Set("cidr_notation", fmt.Sprintf("%s/%d", reservedBlock.Network, reservedBlock.CIDR))
-	return nil
 
+	err = setMap(d, map[string]interface{}{
```
t0mk, line 194: Assignment is unnecessary here, could be returned directly

```
return setMap(d, ma[string]interface{}{
...
```

t0mk, line 194: Assignment is unnecessary here, could be returned directly

return setMap(d, ma[string]interface{}{
...

t0mk (approved):
@displague

I think a normal err check for the d.Sets would be enough, but this makes the code shorter and moved err checks and attr assignments to once place I guess.

It looks good, the tests for reserved IP block succeed. Would be good to have a test to see that the error handling indeed works, but as the checks are done for data returned from the API, it's probably not trivial.

packet/resource_packet_reserved_ip_block.go
displague: Fair enough, I'll pick this up in a follow-up PR that will apply `setMap` more broadly.

golanglint-ci would catch this if https://github.com/packethost/terraform-provider-packet/issues/275 were in place. Once implemented, that will serve as a reminder if I don't pick this up sooner.

PR #280, opened 2020-11-04
device,spot_market_request: add support for custom data

grahamc:
~I haven't actually tried this or even built it, but it is a fairly trivial patch.~ I now run this in production.


grahamc:
Also runs `go mod vendor`, which is useful to me because right now Nixpkgs can only build terraform providers which vendor.

packet/resource_packet_device.go
```diff
@@ -217,6 +217,13 @@ func resourcePacketDevice() *schema.Resource {
 				ForceNew:  true,
 			},
 
+			"custom_data": {
+				Type:      schema.TypeString,
+				Optional:  true,
+				Sensitive: true,
+				ForceNew:  true,
```
displague, line 224: Thanks for the PR, @grahamc!

Could you provide some explanation to justify Sensitive and ForceNew for this field?
I could see a counter-argument that the CustomData should be treated like a data-bus to running instances could use to get access to changing features of the overall Terraform module. 

A user could update CustomData with the certificate or credentials for other services that Terraform is responsible for provisioning, and these values could change independent of the instance.

This has the potential repeat https://github.com/packethost/terraform-provider-packet/issues/236

grahamc: Sure. I don't have any justification other than I assumed it would be best if it behaved similarly to user_data. I wonder if Terraform has a way to let a user define its value to be sensitive on their own, or if there is a way to let the user mark the field as force_new.

displague: That sounds like a use-case for a new type of https://www.terraform.io/docs/configuration/resources.html#lifecycle-lifecycle-customizations

grahamc: Overall though, I have absolutely no opinion on what is the right choice here and would be happy to define it however you'd prefer.

displague: @t0mk what are your thoughts on the flags for this field?

displague: I can justify `Sensitive` in that this field is unbound user data that could express any kind of data, so we error on the sensitive side.

t0mk: @displague I agree that this can be expected to bounce back and forth just like Userdata ForceNew. 

I don't know the purpose of custom data, it's not in the Equinix Console. What's it used for?

t0mk: > That sounds like a use-case for a new type of https://www.terraform.io/docs/configuration/resources.html#lifecycle-lifecycle-customizations (`force_new = [user_data, custom_data]`)

You meant `ignore_changes = [user_data, custom_data]` ? That assumes that both of those attrs could-not-be ForceNew, right?

grahamc: Custom data is not used by any of the Equinix internal systems, but is an additional side-channel like user data for delivering configuration information to the hosts.

displague: > ignore_changes = [user_data, custom_data] 

Not quite. I think that a user should be able to give additional fields the `ForceNew` behavior.  I'm considering opening a TF issue for this type of behavior.  It would help users that expect different behavior from our UserData field too.

displague (approved):

t0mk:
@displague do we want to update the dependencies so aggressively on this kind of small change?

t0mk:
whoops, you've just merged :D

grahamc:
:partying_face: Thank you so much. I'm glad the dependency update made it, too, but understand you'll be removing them soon -- maybe even in the next commit.

displague:
@t0mk this was a `go mod vendor` update that we should have included previously (root go.mod/go.sum was not changed) .
 
I checked it out locally and ran `go mod vendor` just to look for any drift. Oddly enough, there is one file that drifts (`vendor/github.com/hashicorp/hcl/Makefile`) due to `\r` newlines.

We haven't noticed it because we typically are not building with the vendor directory.

The one merge-regret I have is that `custom_data` wasn't included in the creation tests.

displague:
We may want to include `rm -rf vendor; go mod vendor; git diff vendor` as another CI check.

PR #282, opened 2020-11-10
packngo v0.5.0

displague:
By updating to packngo v0.5.0, the provider will make use of the new
api.equinix.com/metal/v1 URLs (rather than api.packet.net).

The network type detection should be improved, as a result, the network
type conversion should be improved.

Devices listing can now use the `search` parameter, which should improve
the performance of hostname lookups in the `packet_device` data source.


displague:
Closing this to re-open it on the root org.

PR #283, opened 2020-11-10
packngo v0.5.0

displague:
By updating to packngo v0.5.0, the provider will make use of the new
api.equinix.com/metal/v1 URLs (rather than api.packet.net).

The network type detection should be improved, as a result, the network
type conversion should be improved.

Devices listing can now use the `search` parameter, which should improve
the performance of hostname lookups in the `packet_device` data source.

### DIFF
--- a/packet/datasource_packet_device.go
+++ b/packet/datasource_packet_device.go
@@ -188,7 +188,7 @@ func dataSourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error 
 		hostname := hostnameRaw.(string)
 		projectId := projectIdRaw.(string)
 
-		ds, _, err := client.Devices.List(projectId, nil)
+		ds, _, err := client.Devices.List(projectId, &packngo.ListOptions{Search: hostname})
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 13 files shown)


detiber (approved):
lgtm

packet/datasource_packet_device.go
```diff
@@ -188,7 +188,7 @@ func dataSourcePacketDeviceRead(d *schema.ResourceData, meta interface{}) error
 		hostname := hostnameRaw.(string)
 		projectId := projectIdRaw.(string)
 
-		ds, _, err := client.Devices.List(projectId, nil)
+		ds, _, err := client.Devices.List(projectId, &packngo.ListOptions{Search: hostname})
```
t0mk, line 191: I was about to write that the hostname attr is optional in this datasource, but I can see that in this branch it should be supplied.

Suggestiong: It might be worth to test if hostname is non-empty before this call, in order to future-proof.

t0mk (commented):
Looks good. I will run a couple of tests with the network type tomorrow. I will approve then.

packet/datasource_packet_device.go
displague: GetOk should guard against empty here: https://godoc.org/github.com/hashicorp/terraform/helper/schema#ResourceData.GetOk

displague:
@t0mk the only tests that failed CI are the usual 1-3 tests

packet/datasource_packet_device.go
t0mk: :+1:

t0mk:
The sweeper error is not good. It's really strange that device listing returns 404... Hopefully it's removed on the next acc test run in the preceeding sweeper.

Otherwise the test fails seem flaky. `TestAccPacketPortVlanAttachment_HybridBasic` succeeded when I ran it with my account.

t0mk (approved):

PR #284, opened 2020-11-10
Add "type" attr to datasoutce precreated_ip_block

t0mk:
Signed-off-by: Tomas Karasek <tom.to.the.k@gmail.com>

### DIFF
--- a/packet/datasource_packet_precreated_ip_block.go
+++ b/packet/datasource_packet_precreated_ip_block.go
@@ -40,6 +40,10 @@ func dataSourcePacketPreCreatedIPBlock() *schema.Resource {
 		Type:     schema.TypeInt,
 		Computed: true,
 	}
+	s["type"] = &schema.Schema{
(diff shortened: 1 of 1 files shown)


t0mk:
@displague 

This is to fix acceptance test run 

```
--- FAIL: TestAccPacketPreCreatedIPBlock_Basic (161.86s)
    testing.go:569: Step 0 error: errors during apply:
        Error: 1 error occurred:
        	* Invalid address to set: []string{"type"}
          on /tmp/tf-test343054386/main.tf line 16:
          (source code not available)
```

displague:
This bug is embarrassingly obvious :-) I definitely thought we were calling `d.Set` with an invalid data type.

We had more flakes than usual in this test run, but the test you were fixing passed. 
I compared against another open PR, and the tests that flaked here are passing there https://github.com/packethost/terraform-provider-packet/pull/283/checks?check_run_id=1380588491

Issue #289, opened 2020-11-16
Unsetting "hardware_reservation_id" does not trigger device re-creation


invidian:
### Terraform Version
```
$ terraform -v
Terraform v0.13.5
+ provider registry.terraform.io/packethost/packet v3.1.0
```

### Affected Resource(s)
- `packet_device`

### Terraform Configuration Files
```diff
terraform {
  required_providers {
    packet = {
      source  = "packethost/packet"
      version = "~> 3.1.0"
    }
  }
}

resource "packet_device" "main" {
  hostname = "mat-test-reservation-id"
-  hardware_reservation_id          = "bar"
+ //hardware_reservation_id          = "bar"
  project_id                       = "foo"
  facilities                       = ["ams1"]
  operating_system                 = "ubuntu_16_04"
  plan                             = "c2.medium.x86"
  billing_cycle                    = "hourly"
  wait_for_reservation_deprovision = true
}

```

### Expected Behavior

When hardware reservation ID is no longer set, device should be re-created with on-demand instance.

### Actual Behavior

Terraform plan is empty and device is not re-created.

### Steps to Reproduce
1. `terraform apply`
2. Apply config change as noted in the diff above.
3. `terraform plan`

### References
Are there any other GitHub issues (open or closed) or Pull Requests that should be linked here? For example:
- kinvolk/lokomotive#1156

invidian:
I ended up with panicking SDK while testing this: https://github.com/hashicorp/terraform-plugin-sdk/issues/642

I _think_ this issue comes from the fact that combination of `Optional`, `Computed` and `ForceNew` is used, which cannot really be well handled from Terraform SDK point of view. One solution would be to split `hardware_reservation_id` field into one which is user-configured and other which is a computed attribute, which can be read e.g. when `next-available` is used.

t0mk:
@invidian thanks for reporting, I will take a look tomorrow. I know the hw res id attr is problematic. It also needs to support the `next-available` value, which makes it a bit hard to interpret, use, and process in the code.

t0mk:
@invidian I did try your use case, and it's true, if you create a device resource with hw res, and then remove the hw res attr, the device is not marked for re-creation.

[...]

PR #293, opened 2020-11-18
Improved handling of hardware reservation ID in the device resource

t0mk:
This PR adds Computed (aka read-only) `deployed_hardware_reservation_id` attribute to the device resource. It can be used to read hw reservation ID when using the "next-available" for picking hw reservation.

It also removes the Computed flag from input attribute `hardware_reservation_id`, fixing #289.


metal/resource_metal_device.go
```diff
@@ -235,6 +235,26 @@ func resourceMetalDevice() *schema.Resource {
...
+			},
+
+			"hardware_reservation_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					dhwr, ok := d.GetOk("deployed_hardware_reservation_id")
+					if ok {
+						if dhwr == new {
```
displague, line 250: what would you expect/want if you changed,
 * from: `hardware_reservation_id=123`
 * to: `hardware_reservation_id=next-available`

displague: I'm wondering if we should suppress diff when the deployed_hardware_reservation_id is set and the new hardware_reservation_id value is "next-available".

t0mk: > What would you expect/want if you changed,
> 
>     * from: `hardware_reservation_id=123`
> 
>     * to: `hardware_reservation_id=next-available`

I understand the logic, and did it on purpose - I myself would expect it to recreate the device. I think we could come up with use cases where a user would expect one way or the other. For me the act (and semantics) of changing an attribute to value "next-available" implies the action of picking a next available slot. If it would be "any" like with the facilities, I'd suggest otherwise.

docs/resources/device.md
```diff
@@ -180,6 +180,7 @@ The following attributes are exported:
 * `billing_cycle` - The billing cycle of the device (monthly or hourly)
 * `created` - The timestamp for when the device was created
 * `deployed_facility` - The facility where the device is deployed.
+* `deployed_hardware_reservation_id` - ID of hw reservation where this device was deployed. It is useful when using the `next-available` hw reservation.
```
displague, line 183: Let's just say "hardware" instead of "hw"

metal/resource_metal_device.go
displague: I just want to make sure we are explicit about the intended behavior.

docs/resources/device.md
```diff
@@ -180,6 +180,7 @@ The following attributes are exported:
...
 * `description` - Description string for the device
 * `hardware_reservation_id` - The ID of hardware reservation which this device occupies
```
displague, line 185: Let's mention the "next-available" value, and what effect changing to this value will have.

displague: We should remove hardware_reservation_id from the attributes reference, it is only an argument now.

displague: "next-available" is documented in the argument reference, so nevermind my first comment.

t0mk: OK! I added some more desc about next-available and removed hw_res_id from the attr list.

t0mk:
@displague I did the fixes to addres your comments.

metal/resource_metal_device.go
invidian: I'd also expect it to re-create the device, it seems logical. And if used wants to suppress this, `ignore_changes` can be used.

docs/resources/device.md
```diff
@@ -150,6 +150,7 @@ The following arguments are supported:
 * `always_pxe` (Optional) - If true, a device with OS `custom_ipxe` will
   continue to boot via iPXE on reboots.
 * `hardware_reservation_id` (Optional) - The `full ID` of the hardware reservation where you want this device deployed, or `next-available` if you want to pick your next available reservation automatically.
+ Changing this from a reservation UUID to `next-available` will re-create the device in another reservation.
```
invidian, line 153:
```suggestion
 Changing this from a reservation UUID to `next-available` will re-create the device with another reservation.
```

invidian (commented):
Some nits, overall looks good. Thanks for taking care of this @t0mk!

docs/resources/device.md
```diff
@@ -150,7 +150,8 @@ The following arguments are supported:
 * `always_pxe` (Optional) - If true, a device with OS `custom_ipxe` will
   continue to boot via iPXE on reboots.
 * `hardware_reservation_id` (Optional) - The `full ID` of the hardware reservation where you want this device deployed, or `next-available` if you want to pick your next available reservation automatically.
-  Please be careful when using hw reservation UUID and `next-available` together for the same pool of reservations. It might happen that the reservation which Equinix Metal API will pick as `next-available` is the reservation which you refer with UUID in another metal_device resource. If that happens, and the metal_device with the UUID is created later, resource creation will fail because the reservation is already in use (by the resource created with `next-available`). To workaround this, have the `next-available` resource  [explicitly depend_on](https://learn.hashicorp.com/terraform/getting-started/dependencies.html#implicit-and-explicit-dependencies) the resource with hw reservation UUID, so that the latter is created first. For more details, see [issue #176](https://github.com/equinix/terraform-provider-equinix-metal/issues/176).
```
displague, line 153: How did metal_device get in here already?

displague: I see. This PR is against `equinix-metal`. Ok.

metal/resource_metal_device.go
```diff
@@ -463,7 +483,7 @@ func resourceMetalDeviceRead(d *schema.ResourceData, meta interface{}) error {
 	}
 
 	if len(device.HardwareReservation.Href) > 0 {
-		d.Set("hardware_reservation_id", path.Base(device.HardwareReservation.Href))
+		d.Set("deployed_hardware_reservation_id", path.Base(device.HardwareReservation.Href))
```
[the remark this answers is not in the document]
displague: We should include that in the CHANGELOG.

This PR is against the `equinix-metal` branch which will be introduced as a new provider, so there are technically no breaking changes 😄, but we do expect users to be able to `sed` their way from `packet` (v3.1.0) to `metal` (v4.0.0), so It will be important to note this change.

metal/resource_metal_device.go
```diff
@@ -235,6 +235,26 @@ func resourceMetalDevice() *schema.Resource {
...
+				Computed: true,
+			},
+
+			"hardware_reservation_id": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					dhwr, ok := d.GetOk("deployed_hardware_reservation_id")
+					if ok {
```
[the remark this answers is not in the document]
t0mk: @invidian we don't have a hw reservation for the testing account right now.

@displague can we have a hw reservation in the testing project? We can either hardcode the ID to the tests (it will be useless for other accounts besides the owner), or we could pass it by envvar. We would also need a dedicated project, I think you can only create a hw res in a project.

docs/resources/device.md
t0mk: I think locative is more proper here. The hw reservation is an allocation slot, in which the device API resource is created.

docs/resources/device.md
t0mk: @displague Is that alright? I can rebase it to another branch.

displague: Fine by me

displague:
This diff has been applied by rebasing the equinix-metal branch.

Thanks, @t0mk.

Issue #298, opened 2020-12-02
packet_reserved_ip_block should include tags (and other missing fields)


displague:
The `packet_reserved_ip_block` resource does not expose the `tag` property (and likely several others) which packngo does include: https://github.com/packethost/packngo/blob/master/ip.go#L41-L70

https://github.com/packethost/terraform-provider-packet/blob/1da99830f51b74dae98661f612f9203adbcf7038/packet/resource_packet_reserved_ip_block.go

displague:
Closing in favor of https://github.com/equinix/terraform-provider-metal/issues/128

PR #300, opened 2020-12-03
Add tags parameter to packet_reserved_ip_block

t0mk:
This PR adds "tags" param to packet_reserved_ip_block.

Signed-off-by: Tomas Karasek <tom.to.the.k@gmail.com>

### DIFF
--- a/packet/resource_packet_reserved_ip_block.go
+++ b/packet/resource_packet_reserved_ip_block.go
@@ -202,8 +202,7 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
 		}
 		quantity = 1 << uint(bits)
 	}
-
-	err = setMap(d, map[string]interface{}{
+	attributeMap := map[string]interface{}{
 		"address": reservedBlock.Address,
 		"facility": func(d *schema.ResourceData, k string) error {
 			if reservedBlock.Facility == nil {
@@ -224,7 +223,15 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
 		"quantity":       quantity,
 		"project_id":     path.Base(reservedBlock.Project.Href),
 		"cidr_notation":  fmt.Sprintf("%s/%d", reservedBlock.Network, reservedBlock.CIDR),
-	})
+	}
+
+	for k := range attributeMap {
+		if d.Get(k) == nil {
+			delete(attributeMap, k)
+		}
+	}
+
+	err = setMap(d, attributeMap)
 	return err
 }
 
--- a/packet/resource_packet_reserved_ip_block.go
+++ b/packet/resource_packet_reserved_ip_block.go
@@ -225,6 +225,7 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
 		"cidr_notation":  fmt.Sprintf("%s/%d", reservedBlock.Network, reservedBlock.CIDR),
 	}
 
+	// filter out attibutes which are not defined in target resource
 	for k := range attributeMap {
 		if d.Get(k) == nil {
 			delete(attributeMap, k)
--- a/packet/resource_packet_reserved_ip_block.go
+++ b/packet/resource_packet_reserved_ip_block.go
@@ -97,6 +97,12 @@ func resourcePacketReservedIPBlock() *schema.Resource {
 		Type:     schema.TypeString,
 		Computed: true,
 	}
+	reservedBlockSchema["tags"] = &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		ForceNew: true,
+		Elem:     &schema.Schema{Type: schema.TypeString},
+	}
 
 	return &schema.Resource{
 		Create: resourcePacketReservedIPBlockCreate,
@@ -134,6 +140,11 @@ func resourcePacketReservedIPBlockCreate(d *schema.ResourceData, meta interface{
 	}
 
 	projectID := d.Get("project_id").(string)
+
+	tags := d.Get("tags.#").(int)
+	if tags > 0 {
+		req.Tags = convertStringArr(d.Get("tags").([]interface{}))
+	}
 
 	blockAddr, _, err := client.ProjectIPs.Request(projectID, &req)
 	if err != nil {
@@ -205,6 +216,7 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
 		"address_family": reservedBlock.AddressFamily,
 		"cidr":           reservedBlock.CIDR,
 		"type":           typ,
+		"tags":           reservedBlock.Tags,
 		"public":         reservedBlock.Public,
 		"management":     reservedBlock.Management,
 		"manageable":     reservedBlock.Manageable,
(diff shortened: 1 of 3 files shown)


t0mk:
@displague this is to partially fix #298

packet/resource_packet_reserved_ip_block.go
```diff
@@ -97,6 +97,12 @@ func resourcePacketReservedIPBlock() *schema.Resource {
 		Type:     schema.TypeString,
 		Computed: true,
 	}
+	reservedBlockSchema["tags"] = &schema.Schema{
+		Type:     schema.TypeList,
+		Optional: true,
+		ForceNew: true,
```
displague, line 103: Tags should be updatable without deleting the reservation (which would result in a different set of IP addresses each time you change a tag).

Is this intentional? Is there a limitation in the API?

packet/resource_packet_reserved_ip_block.go
```diff
@@ -97,6 +97,12 @@ func resourcePacketReservedIPBlock() *schema.Resource {
 		Type:     schema.TypeString,
 		Computed: true,
 	}
+	reservedBlockSchema["tags"] = &schema.Schema{
+		Type:     schema.TypeList,
```
displague, line 101: Should this be `schema.TypeSet`?
 https://www.terraform.io/docs/extend/schemas/schema-types.html#typeset

```
			"tags": {
				Type:        schema.TypeSet,
				Elem:        &schema.Schema{Type: schema.TypeString},
				Optional:    true,
				Description: "An set of tags applied to this reservation.",
			},
```

I'm asking because I'm not sure. I know I have implemented Tags as a TypeSet in the past, https://github.com/linode/terraform-provider-linode/blob/master/linode/resource_linode_instance.go#L109-L110, and had to do some terraform juggling to get the value ready for device creation, comparisons, and updates.

packet/resource_packet_reserved_ip_block.go
t0mk: I put there ForceNew to avoid the need to write resourcePacketReservedIPBlockUpdate (which is nonexistent right now).

I considered the use-case of updating tags of an IP reservation marginal enough to nelect. Would you like me to do the update of the tags?

packet/resource_packet_reserved_ip_block.go
t0mk: A set makes more sense in case the order of the tags is changed - that probably shouldn't be considered a diff. I will make it a set.

Thanks for the heads up with the linode code.

packet/resource_packet_reserved_ip_block.go
displague: If Terraform states that a value should be a certain way, I feel it should make best-effort to represent that state in the host API.

Tags are a value that users of the console.equinix.com UI may manipulate, I think in those cases, Terraform should put the TF defined Tag value back in place.

If other tools, like the CCM or ClusterAPI are manipulating Tags, then it would be best for Terraform to ignore the Tags, which the user can effect with lifecycle configurations.

t0mk: > Tags are a value that users of the console.equinix.com UI may manipulate

There's no way to define tags for the reserved IP block in the Console.
obrazek

.. I've also just created an IP block with the tags and the console didn't even display them, so updating tags over the Console is out of question.

But ok, I'll do an update function for tags and description.

t0mk: @displague I am not sure if it's even possible to update a reservation over the API, it's not documented here: https://metal.equinix.com/developers/api/ipaddresses/, and no PATCH in the /projects/{id}/ips" in the generated api-doc JSON.

displague: Thanks for the extra scrutiny, @t0mk.

displague:
@t0mk the new test is among the failing tests:
```
--- FAIL: TestAccPacketPreCreatedIPBlock_Basic (129.10s)
    testing.go:569: Step 0 error: errors during apply:
        
        Error: 1 error occurred:
        	* Invalid address to set: []string{"tags"}
        
        
        
          on /tmp/tf-test913297341/main.tf line 16:
          (source code not available)
```

t0mk:
@displague ah, thanks. It was because the reserved and pre-created IPs are using the same logic to extract them to TF resource data.

I tried to fix it in a general way, but I'm not sure if you will like it.

packet/resource_packet_reserved_ip_block.go
```diff
@@ -205,13 +216,22 @@ func loadBlock(d *schema.ResourceData, reservedBlock *packngo.IPAddressReservati
...
 		"manageable":     reservedBlock.Manageable,
 		"quantity":       quantity,
 		"project_id":     path.Base(reservedBlock.Project.Href),
 		"cidr_notation":  fmt.Sprintf("%s/%d", reservedBlock.Network, reservedBlock.CIDR),
-	})
+	}
+
+	for k := range attributeMap {
+		if d.Get(k) == nil {
+			delete(attributeMap, k)
```
displague, line 230: I don't understand how this helps.  If the current TF state value is `nil`, we will ignore any API values that would replace `nil`?

displague: The tests passed, which makes me wonder if we are missing some worthwhile tests.

t0mk: [The d.Get() will return nil if the resource doesn't have the attribute defined](https://godoc.org/github.com/hashicorp/terraform-plugin-sdk/helper/schema#ResourceData.Get).

This will filter out all the map items for attributes which are not defined in the target resource, avoiding the `* Invalid address to set: []string{"tags"}`

displague: > If the key does exist in the schema but doesn't exist in the configuration, then the default value for that type will be returned. For strings, this is "", for numbers it is 0, etc.

As long as `nil` is not the default value for any type, this sounds good. I was worried that `nil` would be returned for optional fields.

displague: Looking at https://github.com/hashicorp/terraform/issues/5108#issuecomment-216443491

I think the problem here is that the TypeSet is not willing to take a `["string"]` value, I think we may have to convert the tags to (something like) `map[string]string{"0": "tag"}` to store it as tags, perhaps using a conversion function.

DO seems to do this:

https://github.com/digitalocean/terraform-provider-digitalocean/blob/b3f3acbc88378f6b656735ae39e22dac59e753af/digitalocean/tags.go#L124-L135

displague: Again, I'm not sure how https://github.com/packethost/terraform-provider-packet/pull/300/files#diff-d9548e89dc420c68f877077eab089bc0a1c4d380ece0ad32858280a8111d94caR96-R97 would be passing if that is the case.

t0mk: @displague in the scheme types, nil is default only for TypeInvalid (not sure what TypeInvalid is used for but I think there's no point to set it in a resource)
https://github.com/hashicorp/terraform-plugin-sdk/blob/99085370a27805c4352748aefc0c0b3698cd72d7/helper/schema/schema.go#L2096

t0mk: https://github.com/hashicorp/terraform/issues/5108#issuecomment-216443491 is from 2016, and one of the last comments says it weas patched.

displague:
@t0mk I've learned that Tags can not be updated on IP Reservation blocks.
https://github.com/equinix/cloud-provider-equinix-metal/issues/44#issuecomment-802839224

t0mk:
@displague OK. Shall I move this PR to the new provider repo?

displague:
Closing this to for development in the Equinix Metal provider.

PR #302, opened 2020-12-09
remove daily cron acceptance tests

displague:
The daily cron is not needed for a few reasons:
* PRs are incoming daily
* The cron schedule regularly bumps into PR builds
* The sweepers stomp over active tests
* The packet provider and equinix-metal provider are on the same
schedule.


t0mk:
I agree to drop scheduled tests.

PR #305, opened 2020-12-10
fix BC break with hardware_reservation_id = next-available handling

### LINKED ISSUE #304: 3.2.0 introduced a breaking change without a 4.0.0 version bump
The solution to [#289](https://github.com/packethost/terraform-provider-packet/issues/289) introduced a breaking change without a major version bump.

Can we remove the 3.2.0 version stamp? If so, how does this affect users that are already using 3.2.0?

How does this affect users with the often suggested setting: `hardware_reservation_id = "next-available"` 

What steps should users take when migrating to 4.0.0?

The CHANGELOG should be updated to reflect any changes and

displague:
Fixes #304

### DIFF
--- a/packet/resource_packet_device.go
+++ b/packet/resource_packet_device.go
@@ -245,6 +245,13 @@ func resourcePacketDevice() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					// backward compatibility for < 3.2.0:
+					// removing hardware_reservation_id did not force recreation
+					// (probably because of Computed:true, removed in 3.2.0)
+					if new == "" {
+						return true
+					}
+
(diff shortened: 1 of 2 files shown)


packet/resource_packet_device.go
```diff
@@ -246,6 +246,21 @@ func resourcePacketDevice() *schema.Resource {
...
+
+					// ignore changes to "next-available" when the state matches
+					// the deployed hardware (for "< 3.2.0" compatibility)
+					if ok && new == "next-available" && dwhr == old {
+						return true
+					}
+
+					// preseve legacy behavior to avoid BC break:
+					// ignore moves from UUID -> next-available
+					if new == "next-available" && len(old) > 0 {
```
displague, line 258: I suspect the rule at L252-254 is sufficient and we could remove L258-260 in 4.0.0.

PR #310, opened 2022-06-09
Add container security scanning

JAORMX:
This adds a workflow that runs Trivy [1] on all Dockerfiles in the repository.

The intention is to verify that Pull Requests don't introduce security vulnerabilities.

The workflow will also run on a specified cadence (Tuesdays at 15:15 UTC). Which
will allow us to discover new package-level vulnerabilities in the container(s).

Note that scanning container images in clusters will be handled elsewhere and not
as part of this PR.

[1] https://aquasecurity.github.io/trivy/

### DIFF
--- a/.github/workflows/container-scan-trivy.yml
+++ b/.github/workflows/container-scan-trivy.yml
@@ -1,0 +1,40 @@
+---
+name: trivy-container-scan
+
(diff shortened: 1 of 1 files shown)


displague:
Not accepting this PR. This project does not create containers, only binaries. Docker is not used for build (GH Actions are).

